PR #3, opened 2016-03-31
Add a ConsumerDaemon

jessemyers:
Depends on [this PR](https://github.com/globality-corp/microcosm-postgres/pull/4)

Adds a consumer daemon base class that reads messages from SQS and dispatches them to user functions.

Must be subclassed to define:
 - The app/graph name
 - The SQS queue URL
 - A mapping from message media type to Schema object
 - A mapping from message media type to user functions

Mesages are dispatched by type to user functions and skipped if there's no matching type. User functions can also skip/ignore messages, though this has no behavior difference presently except some stats/accounting. User functions that raise exceptions will cause the message to not be deleted, meaning that it will be retried and dead-lettered if it continues to fail.

Here's an example:

```
from microcosm_pubsub.daemon import ConsumerDaemon
from microcosm_pubsub.main import SimpleSchema


def handle(message):
    """Example user function"""
    print message
    return True


class SimpleConsumerDaemon(ConsumerDaemon):

    @property
    def name(self):
        return "example"

    @property
    def sqs_queue_url(self):
        return "https://sqs.us-west-2.amazonaws.com/xxx/yyy"

    @property
    def handler_mappings(self):
        return {
            SimpleSchema.MEDIA_TYPE: handle,
        }

    @property
    def schema_mappings(self):
        return {
            SimpleSchema.MEDIA_TYPE: SimpleSchema,
        }


if __name__ == '__main__':
    daemon = SimpleConsumerDaemon()
    daemon.run()
```


cdm-ium:
:+1:  I'd kind of like your detailed example in the readme or a `example` file or something.

jessemyers:
+1

I updated the README and made a few small improvements.

PR #4, opened 2016-04-18
Allow SQS message handlers to be configured from the graph.

jessemyers:
My earlier attempt at wiring in SQS message handlers turned out to occur before the graph was even created, so I'm doing the right thing and making it possible for the handler mapping to be a graph component.


amitbn:
+1

PR #5, opened 2016-04-25
Tune SQS consumer behavior by setting a wait timeout of 1s.

jessemyers:
Setting to zero (the previous behavior) fails to wait long enough for new messages to appear;
this causes long delays on message delivery. Setting to larger values causes the daemon to block
waiting on SQS; this interferes with daemon interruption and shutdown.

At the same time: add the simple daemon from the README to the examples and modify the example
message to better enable diagnostics.

### DIFF
--- a/microcosm_pubsub/consumer.py
+++ b/microcosm_pubsub/consumer.py
@@ -147,7 +147,7 @@ class SQSConsumer(object):
     # SQS will not return more than ten messages at a time
     limit=10,
     # SQS will only return a few messages at time unless long polling is enabled (>0)
-    wait_seconds=0,
+    wait_seconds=1,
 )
 def configure_sqs_consumer(graph):
     """
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -6,6 +6,7 @@ from collections import namedtuple
 from logging import getLogger
 
 from microcosm.api import defaults
+from microcosm.errors import NotBoundError
 
 
 DispatchResult = namedtuple("DispatchResult", ["message_count", "error_count", "ignore_count"])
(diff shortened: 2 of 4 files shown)


microcosm_pubsub/consumer.py
```diff
@@ -147,7 +147,7 @@ def nack(self, message):
     # SQS will not return more than ten messages at a time
     limit=10,
     # SQS will only return a few messages at time unless long polling is enabled (>0)
-    wait_seconds=0,
+    wait_seconds=1,
```
jessemyers: This is the main change. The rest is just improvements around testing.

microcosm_pubsub/dispatcher.py
```diff
@@ -69,7 +70,7 @@ def configure_sqs_message_dispatcher(graph):
     """
     try:
         sqs_message_handlers = graph.sqs_message_handlers
-    except AttributeError:
+    except NotBoundError:
```
jessemyers: This was a bug: use of the daemon `schema_mappings` property wasn't working since failure to supply the `sqs_message_handlers` binding would raise an uncaught exception.

microcosm_pubsub/main.py
```diff
@@ -18,6 +21,7 @@ class SimpleSchema(PubSubMessageSchema):
     MEDIA_TYPE = "application/vnd.globality.pubsub.simple"
 
     message = fields.String(required=True)
+    timestamp = fields.Float(required=True)
```
jessemyers: I modified the example schema to have a timestamp so we can easily trace delivery time.

microcosm_pubsub/main.py
```diff
@@ -72,3 +80,39 @@ def load_config(metadata):
     for message in messages:
         with message:
             print message.content  # noqa
+
+
+class SimpleConsumerDaemon(ConsumerDaemon):
```
jessemyers: This is basically the example from the README with some timing information printed on consume.

setup.py
```diff
@@ -29,6 +29,7 @@
         "console_scripts": [
             "sns-produce = microcosm_pubsub.main:produce",
             "sqs-consume = microcosm_pubsub.main:consume",
+            "simple-daemon = microcosm_pubsub.main:main",
```
jessemyers: I used this daemon to test different wait settings.

PR #6, opened 2016-04-27
uuid field type + its tests

amitbn:
We want to use the this new `UUIDField` in schemas which are using UUID fields (instead of marshmallow's given fields.UUID). The reason for that is that `UUID` type is not json-serializable so pass it through string conversion first.

### DIFF
--- a/microcosm_pubsub/fields.py
+++ b/microcosm_pubsub/fields.py
@@ -10,10 +10,6 @@ class UUIDField(Field):
     UUID valued field, as either a string or UUID object.
 
     """
-    def __init__(self, use_uuidformat=False, *args, **kwargs):
-        self.use_uuidformat = use_uuidformat
-        super(UUIDField, self).__init__(*args, **kwargs)
-
     def _serialize(self, value, attr, obj):
         """
         Serialize value as a uuid, either as a string or an UUID object.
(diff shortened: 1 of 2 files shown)


microcosm_pubsub/fields.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        Deserialize value as a uuid.
+
+        Handle both string and UUID object.
+        """
+        if value is None:
+            return None
+
+        try:
+            if self.use_uuidformat:
```
jessemyers: I feel like this clause should check whether the value is an instance of a UUID and do the right thing rather than using the boolean.

amitbn: Don't you think that using `instanceof` is a bit messy here?
Can you please explain why is it better practice than passing the boolean flag `use_uuidformat` when initializing?

jessemyers: You don't have control of your inputs in all cases. It's equally plausible that you are passed a `UUID` or a string. But no matter what you are passed, we want to return a string so that the pubsub layer doesn't blow up. Right?

amitbn:
@jessemyers - back to you after changing what's required

jessemyers:
LGTM

PR #7, opened 2016-05-02
Relax pubsub configuration assumptions

jessemyers:
- Queue URL may not be provided on the CLI
 - Handler mappings may not be configured as a property

### DIFF
--- a/microcosm_pubsub/daemon.py
+++ b/microcosm_pubsub/daemon.py
@@ -12,7 +12,7 @@ class ConsumerDaemon(Daemon):
 
     def make_arg_parser(self):
         parser = super(ConsumerDaemon, self).make_arg_parser()
-        parser.add_argument("--sqs-queue-url", required=True)
+        parser.add_argument("--sqs-queue-url")
         return parser
 
     @abstractproperty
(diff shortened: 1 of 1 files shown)


microcosm_pubsub/daemon.py
```diff
@@ -12,7 +12,7 @@ class ConsumerDaemon(Daemon):
 
     def make_arg_parser(self):
         parser = super(ConsumerDaemon, self).make_arg_parser()
-        parser.add_argument("--sqs-queue-url", required=True)
+        parser.add_argument("--sqs-queue-url")
```
jessemyers: I'm specifically changing my daemon to have a reasonable default for the queue URL; it makes local development configuration a lot simpler.

microcosm_pubsub/daemon.py
```diff
@@ -23,27 +23,37 @@ def schema_mappings(self):
         """
         pass
 
-    @abstractproperty
     def handler_mappings(self):
         """
         Define the PubSub message media-type to handler mappings.
 
+        This function is an alternative to configuring `graph.sqs_message_handlers` as
```
jessemyers: Most of our daemons have stopped using this property; I'm not yet ready to remove it, but I am making it less required.

microcosm_pubsub/daemon.py
amitbn: makes sense

amitbn:
+1

PR #8, opened 2016-06-01
Fix sns typo

pokey:


adamhadani:
LGTM , @jessemyers  double check me here for sanity? a daemon consumes from an *SQS* queue, (whereas we publish to an SNS topic), so this argument should be an sqs ARN/URL?

jessemyers:
@adamhadani, yes write to SNS, read from SQS. The former takes an ARN, the latter a queue URL.

PR #10, opened 2016-07-15
allow for binding a default handler for malformatted messages

cdm-ium:
I want to be able to register a `default` handler which can be used when no handler for a desired media type can be found or you *have* to parse a message with no media type.

### DIFF
--- a/microcosm_pubsub/consumer.py
+++ b/microcosm_pubsub/consumer.py
@@ -6,6 +6,7 @@ from json import loads
 from hashlib import md5
 
 from boto3 import client
+from marshmallow.exceptions import ValidationError
 from microcosm.api import defaults
 
 
@@ -72,7 +73,11 @@ class SQSMessage(object):
         Parse and validate the message.
 
         """
-        base_message = consumer.pubsub_message_codecs["_"].decode(message)
+        try:
+            base_message = consumer.pubsub_message_codecs["_"].decode(message)
+        except ValidationError:
+            return loads(message)
+
         media_type = base_message["mediaType"]
         content = consumer.pubsub_message_codecs[media_type].decode(message)
         return content
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -48,7 +48,7 @@ class SQSMessageDispatcher(object):
         Handle a single message.
 
         """
-        media_type = message["media_type"]
+        media_type = message.get("media_type", "_")
         sqs_message_handler = self.sqs_message_handlers.get(media_type)
         if sqs_message_handler is None:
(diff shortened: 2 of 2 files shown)


microcosm_pubsub/dispatcher.py
```diff
@@ -48,7 +48,7 @@ def handle_message(self, message):
         Handle a single message.
 
         """
-        media_type = message["media_type"]
+        media_type = message.get("media_type", "_")
```
cdm-ium: This allows for choosing a `default` handler registered on the `_` key.

microcosm_pubsub/consumer.py
```diff
@@ -72,7 +73,11 @@ def parse_content(cls, consumer, message):
         Parse and validate the message.
 
         """
-        base_message = consumer.pubsub_message_codecs["_"].decode(message)
+        try:
```
jessemyers: As I mentioned in [https://github.com/globality-corp/oscar/pull/158](), I think we should move the initial parsing logic into a separate construct so it can be overridden more simply.

cdm-ium: Is the idea, that these functions `parse_content` and `handle_message` should be factored out into something with a more easily overriden API so they can be overridden with logic which doesn't assume your messages have a `mediaType`.

microcosm_pubsub/dispatcher.py
cdm-ium: Are you OK with this part? Perhaps rather than explicitly using `_` we can add a parameter to the daemon for `microcosm_pubsub.default_key` and use that value so it is clearer what the key is for.

microcosm_pubsub/consumer.py
jessemyers: I think my plan would be:

1. Change codec.py to not have a `default` codec and to remove the `_` dict mappings key. It should be opt-in to set a default codec so that it is an error to not configure a codec for a media type. The use of `defaultdict` would need to be conditional on having a `default` value configured.

2. The consumer should have an "envelope decoding" function; this should default to the current behavior, using the `MediaType` schema; it should be possible to wire in an alternate decoding function by setting an appropriate object graph key.

 3. The simple schema and CLI produce/consume should probably be deleted or moved to an example directory (and removed from the README); they only add confusion because people expect them to work with custom schemas and they don't.

microcosm_pubsub/dispatcher.py
jessemyers: It probably makes more sense for `SQSMessage.from_sqs` to return a `(media_type, message)` tuple and have both args get passed here. It would be the responsibility of the "envelope decoding" function to extract this media type; in this case, you'd inject a function that returned: `"_", loads(message)`

PR #11, opened 2016-07-17
Refactor the envelope parsing from SQS

jessemyers:
Main changes:
 - Moved parsing functions to new module instead of classmethods of the SQSMessage object.
 - Use subclasses of `SQSEnvelope` for alternate parsing patterns
 - Explicitly use the MediaTypeSchema instead of depending on `_` naming convention.
 - Add raw parsing of envelope as JSON
 - Remove the default usage of a default schema. (Unrelated, but a good idea.)

### DIFF
--- a/microcosm_pubsub/codecs.py
+++ b/microcosm_pubsub/codecs.py
@@ -91,11 +91,9 @@ class PubSubMessageCodec(object):
 
 
 @defaults(
-    default=PubSubMessageSchema,
+    default=None,
     strict=True,
-    mappings=dict(
-        _=MediaTypeSchema,
-    ),
+    mappings=dict(),
 )
 def configure_pubsub_message_codecs(graph):
     """
@@ -105,7 +103,10 @@ def configure_pubsub_message_codecs(graph):
     default_schema_cls = graph.config.pubsub_message_codecs.default
     strict = graph.config.pubsub_message_codecs.strict
 
-    message_codecs = defaultdict(lambda: PubSubMessageCodec(default_schema_cls(strict=strict)))
+    if default_schema_cls:
+        message_codecs = defaultdict(lambda: PubSubMessageCodec(default_schema_cls(strict=strict)))
+    else:
+        message_codecs = dict()
 
     for key, schema_cls in graph.config.pubsub_message_codecs.mappings.items():
         message_codecs[key] = PubSubMessageCodec(schema_cls(strict=strict))
--- a/microcosm_pubsub/consumer.py
+++ b/microcosm_pubsub/consumer.py
@@ -8,15 +8,18 @@ from hashlib import md5
 from boto3 import client
 from microcosm.api import defaults
 
+from microcosm_pubsub.codecs import MediaTypeSchema, PubSubMessageCodec
+
 
 class SQSMessage(object):
     """
     SQS message wrapper.
 
     """
-    def __init__(self, consumer, content, message_id, receipt_handle):
+    def __init__(self, consumer, content, media_type, message_id, receipt_handle):
         self.consumer = consumer
         self.content = content
+        self.media_type = media_type
         self.message_id = message_id
         self.receipt_handle = receipt_handle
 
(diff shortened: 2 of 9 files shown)


microcosm_pubsub/codecs.py
```diff
@@ -105,7 +103,10 @@ def configure_pubsub_message_codecs(graph):
     default_schema_cls = graph.config.pubsub_message_codecs.default
     strict = graph.config.pubsub_message_codecs.strict
 
-    message_codecs = defaultdict(lambda: PubSubMessageCodec(default_schema_cls(strict=strict)))
+    if default_schema_cls:
```
jessemyers: It's still possible to have a default codec, but in practice this has been a bad idea since it meant that you might publish a message and only have its media type get sent (instead of any of its content). So now you have to opt-in for this behavior.

microcosm_pubsub/consumer.py
```diff
@@ -142,6 +95,56 @@ def nack(self, message):
...
+        """
+        message_id = sqs_message["MessageId"]
+        receipt_handle = sqs_message["ReceiptHandle"]
+        body = sqs_message["Body"]
+
+        if validate_md5:
+            self._validate_md5(sqs_message, body)
+
+        message = loads(body)["Message"]
+        media_type, content = self._parse_message_from_sqs(message)
```
jessemyers: We parse the media type and content here. If we want to support other kind of message enveloping (e.g. using SQS metadata) or bypass enveloping (using a raw message) with a vanilla media type of `application/json`, we just need to swap out this function.

The next step would be to build an "enveloping" module and have this function chose based on config.

microcosm_pubsub/consumer.py
```diff
@@ -142,6 +95,56 @@ def nack(self, message):
...
+        )
+
+    def _parse_message_from_sqs(self, message):
+        """
+        Parse and validate SQS message content.
+
+        :returns: a media_type, content tuple
+
+        """
+        base_message = PubSubMessageCodec(MediaTypeSchema()).decode(message)
```
jessemyers: No more `_` lookup anymore. We're explicit instead.

microcosm_pubsub/consumer.py
```diff
@@ -2,107 +2,24 @@
...
 
 """
-from json import loads
-from hashlib import md5
-
 from boto3 import client
 from microcosm.api import defaults
 
 
-class SQSMessage(object):
```
jessemyers: Moved to its own module for simpler imports. Classmethods removed.

microcosm_pubsub/dispatcher.py
```diff
@@ -35,20 +35,19 @@ def handle_batch(self):
             message_count += 1
             try:
                 with message:
-                    if not self.handle_message(message.content):
+                    if not self.handle_message(message.media_type, message.content):
```
jessemyers: Pass media type here.

microcosm_pubsub/envelope.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        message_id = raw_message["MessageId"]
+        receipt_handle = raw_message["ReceiptHandle"]
+        body = raw_message["Body"]
+
+        if self.validate_md5:
+            self.validate_md5(raw_message, body)
+
+        message = loads(body)["Message"]
+        media_type, content = self.parse_media_type_and_content(message)
```
jessemyers: We now return the parsed media type here and attach it to the message.

microcosm_pubsub/envelope.py
```diff
@@ -0,0 +1,130 @@
...
+            ))
+
+
+class RawSQSEnvelope(SQSEnvelope):
+    """
+    Enveloping strategy that just passes raw JSON.
+
+    """
+    def parse_media_type_and_content(self, message):
+        content = loads(message)
```
jessemyers: I'm hoping this strategy suffices for SES parsing?

cdm-ium: I will test it and report back.

microcosm_pubsub/consumer.py
```diff
@@ -166,8 +83,8 @@ def configure_sqs_consumer(graph):
 
     return SQSConsumer(
```
jessemyers: Yes. But you want that anyway because you should *definitely* be using a different queue for email.

cdm-ium:
+1

PR #12, opened 2016-07-19
adding binding for sqs envelope

cdm-ium:
I think we need a binding for the `sqs_envelope` so it is accessible via import.


jessemyers:
I think you missed the same thing in `setup.py`; you just need to reinstall `microcosm-pubsub`

cdm-ium:
@jessemyers my bad. Please cut a release then. This is working for me!

PR #13, opened 2016-08-17
attaching message context to handler invocations

cdm-ium:
Pubsub equivalent of: https://github.com/globality-corp/microcosm-flask/pull/51

### DIFF
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -21,10 +21,9 @@ class SQSMessageDispatcher(object):
     Dispatch batches of SQSMessages to handler functions.
 
     """
-    def __init__(self, sqs_consumer, sqs_message_handlers, log_with_context, sqs_message_context):
+    def __init__(self, sqs_consumer, sqs_message_handlers, sqs_message_context):
         self.sqs_consumer = sqs_consumer
         self.sqs_message_handlers = sqs_message_handlers
-        self.log_with_context = log_with_context
         self.sqs_message_context = sqs_message_context
 
     def handle_batch(self):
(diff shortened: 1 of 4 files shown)


microcosm_pubsub/dispatcher.py
```diff
@@ -78,4 +91,6 @@ def configure_sqs_message_dispatcher(graph):
     return SQSMessageDispatcher(
         sqs_consumer=graph.sqs_consumer,
         sqs_message_handlers=sqs_message_handlers,
+        log_with_context=graph.config.sqs_message_dispatcher.log_with_context,
```
jessemyers: Seems like if you set log with context to False, you are still required to wire in `graph.message_context`?

microcosm_pubsub/dispatcher.py
```diff
@@ -78,4 +91,6 @@ def configure_sqs_message_dispatcher(graph):
...
+        message_context=graph.message_context,
```
jessemyers: How about `sqs_message_context`?

jessemyers:
LGTM if you address the usability/configurability issues.

microcosm_pubsub/dispatcher.py
```diff
@@ -75,7 +83,15 @@ def configure_sqs_message_dispatcher(graph):
     except NotBoundError:
         sqs_message_handlers = graph.config.sqs_message_dispatcher.mappings
 
+    try:
```
cdm-ium: I removed the `log_with_context` flag in favor of having a noop default that we can always use. User can have more fine-grained control by binding the key `sqs_message_context` with either the default function from this repo or their own custom context.

PR #16, opened 2016-08-24
fixing exception logging

cdm-ium:
This will make sure that daemon exceptions are logged with request id and any other available message metadata.


jessemyers:
LGTM

PR #18, opened 2016-08-25
Usage of opaque data.

cdm-ium:
This is an example usage of: https://github.com/globality-corp/microcosm/pull/12

@jessemyers  I see the value of your comments about api fluency here. For one thing, it would allow me to clean up the mocking of opaque by factories when the graph is instantiated in test mode. I.e. right now I have to do:

```
    if graph.metadata.testing:
        from mock import MagicMock
        opaque = MagicMock(data_func=MagicMock(return_value=dict()))
    else:
        opaque = graph.opaque
```

And with your recommendation I would be able to do:

```
    if graph.metadata.testing:
        from mock import MagicMock
        opaque = dict()
    else:
        opaque = graph.opaque
```


jessemyers:
LGTM subject to fluency improvements

PR #21, opened 2016-08-29
Allow deferring publishing until the end of a context block.

jessemyers:
This lets us do something like:

```
with DeferredProducer(graph.sns_producer) as producer:
    with transaction():
        some_db_change()
        producer.produce(...)
        some_other_db_change()
```

and have the publish occur after the transaction finishes.


cdm-ium:
+1

PR #22, opened 2016-09-01
Make failed message available immediately

srt32:
Why:

* if we have a queue with a visibility timeout of 30 seconds and
  message processing fails immediately, the message will not be
  available for processing for 30 seconds. This kind of window is
  helpful for handling timeouts as we want to give the job a chance to
  succeed but upon fast failures we want to make the message available
  again.

This change addresses the need by:

* upon failure of processing a job, we call `nack`. With this change,
  `nack` will now change the visibility timeout of the failed message to
  `0`, which makes it immediately available. Per the
  [docs](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ChangeMessageVisibility.html)
  (and see snippet below)
  this setting is not persisted across attempts so the next time the
  message is picked up, its visibility timeout will be set back to the
  queue's default. This behavior is desireable because the second
  attempt at processing will start quickly but then have its normal 30
  seconds to succeed.
* Open question: where should some kind of backoff happen for message
  failure? Should it be here or in `microcosm-daemon`?

```
Unlike with a queue, when you change the visibility timeout for a
specific message, that timeout value is applied immediately but is not
saved in memory for that message. If you don't delete a message after it
is received, the visibility timeout for the message the next time it is
received reverts to the original timeout value, not the value you set
with the ChangeMessageVisibility action.
```

### DIFF
--- a/microcosm_pubsub/consumer.py
+++ b/microcosm_pubsub/consumer.py
@@ -11,17 +11,21 @@ class SQSConsumer(object):
     Consume message from a (single) SQS queue.
 
     """
-    def __init__(self,
-                 sqs_client,
-                 sqs_envelope,
-                 sqs_queue_url,
-                 limit,
-                 wait_seconds):
+    def __init__(
+        self,
+        sqs_client,
+        sqs_envelope,
+        sqs_queue_url,
+        limit,
+        wait_seconds,
+        visibility_timeout_seconds,
+    ):
         self.sqs_client = sqs_client
         self.sqs_envelope = sqs_envelope
         self.sqs_queue_url = sqs_queue_url
         self.limit = limit
         self.wait_seconds = wait_seconds
+        self.visibility_timeout_seconds = visibility_timeout_seconds
 
     def consume(self):
         """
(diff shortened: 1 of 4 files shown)


jessemyers:
My suggestions:

 1. Make the `nack` take an argument and set the visibility timeout accordingly. We may want to opt-in to this behavior (or opt-out) instead of always doing it.
```
def nack(retry=None):
```
 2. Investigate what meta data we can get from AWS around the underlying message. AWS keeps track of the retry itself because that's how dead-lettering works. I imagine it's exposed somewhere. If so, we can add another parameter to the `nack` argument to define the scaling factor of the retry and build an actual back off strategy.

adamhadani:
Lgtm
Out of curiosity, what is the reason this specific pubsub message tends to fail fairly often l?

cdm-ium:
I'm guessing this might have unintended consequences. If this significantly speeds up brief availability time it will be great, but my vote is to merge this after the push to test and test in dev so we can be destructive without interrupting people as much. If it looks good, we can promote it.

Thoughts @jessemyers @srt32 ?

jessemyers:
LGTM assuming:
 - We make the behavior opt-in via `nack` argument (so releasing this library is safe)
 - We wait until we've pushed to test (an hour or two) to update daemons to pass a different argument (you'll have to figure out how a deamon controls the `nack` argument)
 - We cherry-pick the changes to the release branch if it looks good.

srt32:
I agree with waiting to merge / update apps.

re: adding a hook to pass in scaling factor and retries, I'll explore our options.

srt32:
Updated in 662406a to add a new config variable that gets passed to the `SQSConsumer` and then used in calling `nack`.

jessemyers:
LGTM

PR #24, opened 2016-09-02
GLOB-3763 - properly set visibility default

srt32:
Why:

* Following on from
  https://github.com/globality-corp/microcosm-pubsub/pull/22 there was a
  missing default for visibilty_timeout_seconds, which would make the
  change non-backwards compatible.

This change addresses the need by:

* adds `visibilty_timeout_seconds` to the defaults for the daemon

https://globality.atlassian.net/browse/GLOB-3763


netama:
LGTM

srt32:
Thanks!

PR #26, opened 2016-09-02
Update develop after 0.13.1 release

srt32:


jessemyers:
We don't need this version bump; we'll bump again when we make our next release.

PR #28, opened 2016-09-02
cast visibility timeout as int of not None

sethisernhagen:
There is an error if visibility_timeout_seconds is a string.  Cast value to int if not None.


netama:
+1

microcosm_pubsub/consumer.py
```diff
@@ -85,6 +85,9 @@ def configure_sqs_consumer(graph):
     wait_seconds = graph.config.sqs_consumer.wait_seconds
     visibility_timeout_seconds = graph.config.sqs_consumer.visibility_timeout_seconds
 
+    if visibility_timeout_seconds:
+        visibility_timeout_seconds = int(visibility_timeout_seconds)
```
jessemyers: You can also use: `foo =  int(bar) if bar is not None else None`

PR #29, opened 2016-10-02
Enable fluent decorators for registering PubSub schemas and handlers.

jessemyers:
This is best demonstrated by example:

    @schema
    class MySchema(PubSubMessageSchema):
        my_field = fields.String()

    @binding("my_handler")
    @handles(MySchema)
    class MyHandler(object):

        def __init__(self, graph):
            self.collaborator = graph.collaborator

        def __call__(self, message):
            self.collaborator.do_something(message)
            return True

The underlying change involves moving the schema and handler mappings into a centralized,
singleton registry. This is a small loss in terms of object graph purity -- these mappings
are stored in singletons, not the object graph because it's rather inconvenient to enforce
that the graph exists at the point in time that these types are declared -- but the benefit
is a great deal more clarity to what was a frequent source of usage errors.

The legacy registry mechanism (namely, dictionaries bound to the graph) still exists, but
is implemented by loading these dictionaries into the registry when the graph binds the
registries. This behavior still needs some testing in anger.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,20 +19,6 @@ PubSub with SNS/SQS
     by separate consumers.
 
 
-## CLI
-
-For testing purposes, the producer and consumer functions can be invoked from the CLI. These test functions
-use a very simplistic schema and will not work for most real use cases.
-
-To produce messages:
-
-    sns-produce --topic-arn <topic-arn>
-
-To consume messages:
-
-    sqs-consume --queue-url <queue-url>
-
-
 ## Validation
 
 Messages use [marshmallow](http://marshmallow.readthedocs.org/en/latest/index.html) schemas for validation.
@@ -81,17 +67,21 @@ no exception is raised during processing:
 ## Asynchronous Workers
 
 The `ConsumerDaemon` base class supports creating asynchronous workers ("daemons") that consume
-messages and dispatch them to user-defined worker functions. Usage requires creating a subclass
-that defines the `name` and `schema_mappings` properties and the `sqs_message_handlers` binding.
+messages and dispatch them to user-defined worker functions. Usage involves declaring a schema,
+declaring a handler function, and declaring a deamon that runs them.
 
 
-Import the baseclass and define a schema.
+Import the baseclass, define a schema, and decorate it with `@schema`:
 
     from marshmallow import fields
 
-    from microcosm_pubsub.daemon import ConsumerDaemon
-    from microcosm.api import create_object_graph
+    from microcosm.api import binding, create_object_graph
 
+    from microcosm_pubsub.daemon import ConsumerDaemon
+    from microcosm_pubsub.decorators import handles, schema
+
+
+    @schema
     class SimpleSchema(PubSubMessageSchema):
         """
         A single schema that just sends a text string.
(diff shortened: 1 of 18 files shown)


microcosm_pubsub/consumer.py
```diff
@@ -76,6 +76,8 @@ def nack(self, message, visibility_timeout_seconds=None):
     limit=10,
     # SQS will only return a few messages at time unless long polling is enabled (>0)
     wait_seconds=1,
+    # By default, don't change message visibility on nack
+    visibility_timeout_seconds=None,
```
jessemyers: Housekeeping: This config was previously defined in the consumer daemon's defaults instead of in the defaults of the relevant component.

microcosm_pubsub/daemon.py
```diff
@@ -16,34 +14,35 @@ def make_arg_parser(self):
         parser.add_argument("--sqs-queue-url")
         return parser
 
-    @abstractproperty
```
jessemyers: We no longer need to define schema mappings here. I haven't removed them yet, but I probably will eventually.

README.md
```diff
@@ -105,34 +95,29 @@ Import the baseclass and define a schema.
...
+handlers will be a class with its own `@binding` to pass other collaborators:
 
-    def handle_simple(message):
-        print message
-        return True
+    @binding("simple_handler")
+    @handles(SimpleSchema)
+    class SimpleHandler(object):
+        def __init__(self, graph):
+            self.collaborator = graph.collaborator
```
amitbn: Can you please explain what's the purpose of the `collaborator`.
And also, in which layer is is being initialized?

amitbn:
That's pretty nice.
So now we won't need these 2 declarations on the `main.py` of a consumer daemon?
1) https://github.com/globality-corp/pilar/blob/develop/pilar/daemon/main.py#L48
2) https://github.com/globality-corp/pilar/blob/develop/pilar/daemon/main.py#L55
?

microcosm_pubsub/daemon.py
```diff
@@ -16,34 +14,36 @@ def make_arg_parser(self):
...
+    def create_object_graph(self, args):
+        graph = super(ConsumerDaemon, self).create_object_graph(args)
+        for media_type, schema_cls in self.schema_mappings.items():
+            self.graph.pubsub_message_schema_registry.register(media_type, schema_cls)
+        return graph
 
     @property
     def defaults(self):
-        dct = dict(
-            pubsub_message_codecs=dict(
```
amitbn: I'm confused. 
Aren't `pubsub_message_codecs` the messages we want to **publish**? How's that related to this PR?

microcosm_pubsub/fields.py
```diff
@@ -1,9 +1,11 @@
 """
 Custom fields.
+
 """
-from marshmallow.fields import Field, ValidationError
 from uuid import UUID
 
+from marshmallow.fields import Field, ValidationError
+
```
amitbn: Out of curiosity since you've changed the order of the imports - does it matter here?

microcosm_pubsub/daemon.py
jessemyers: No, `pubsub_message_codecs` is a translation between `media_type` and `schema`. It's used for both publish and subscribe. Part of why this PR is valuable is exactly this confusion; we frequently had handlers get registered for a `media_type`, but not the accompanying `schema`; this meant that we never decoded the message to pass to the schema and the daemon didn't work. The new `@schema` decorator makes that problem vanish by binding the handler to a schema class.

microcosm_pubsub/fields.py
jessemyers: No, I'm just fixing up a `PEP8` violation; built-ins like `uuid` should come first.

jessemyers:
Correct, we'll be able to simplify the `daemon` to just decorated classes and a single subclass with appropriate imports. (Note that using decorators for wiring without an explict `graph.use` means we need to be careful about import scope if we have multiple daemons.)

microcosm_pubsub/dispatcher.py
```diff
@@ -59,53 +63,25 @@ def handle_message(self, media_type, message):
         if message is None:
             logger.debug("Skipping message with unparsed type: {}".format(media_type))
```
netama: `self.`?

netama: `self.`?

microcosm_pubsub/dispatcher.py
```diff
@@ -59,53 +63,25 @@ def handle_message(self, media_type, message):
...
-@defaults(
-    mappings=dict(),
-)
-def configure_sqs_message_dispatcher(graph):
-    """
-    Configure dispatching of SQS messages.
+            try:
+                sqs_message_handler = self.sqs_message_handler_registry.find(media_type)
+            except KeyError:
+                # no handlers
```
netama: Can this case still have the "skipping" debug log?

microcosm_pubsub/registry.py
```diff
@@ -0,0 +1,127 @@
...
+    def __init__(self, graph):
+        super(PubSubMessageSchemaRegistry, self).__init__(graph)
+        self.strict = graph.config.pubsub_message_schema_registry.strict
+
+    @property
+    def legacy_binding_key(self):
+        return "pubsub_message_codecs"
+
+    @property
+    def allows_multiple(self):
```
netama: where is this used? should it also be in the base?

netama:
+1 (or 10)

microcosm_pubsub/dispatcher.py
jessemyers: 👍

microcosm_pubsub/registry.py
jessemyers: It's an artifact of an earlier design. I'll delete it.

PR #30, opened 2016-10-06
Eliminate singleton by instead using object graph hooks.

jessemyers:
Remove singletons and instead use [deferred hooks](https://github.com/globality-corp/microcosm/pull/15)

Requires a new `microcosm` release to pass builds.


netama:
LGTM

PR #31, opened 2016-10-07
Add missing argument

afallou:


jessemyers:
LGTM

Shouldn't be a problem in practice since Python doesn't enforce uniform signatures on subclass functions (which is why it wasn't caught before), but useful as documentation.

PR #32, opened 2016-10-11
Convention-driven pubsub messaging.

jessemyers:
In a (immutable) event-driven system, pubsub messages are just reflections of state changes made
in some source of truth, typically a CRUD-oriented microservice. (See `microcosm-flask`).

These state changes can be represented very simply as a type and URI, plus some optional opaque
data for accounting/logging purposes. During a normal message handling flow, the consumer queries
back to the source of truth (by URI) and fetches further information. Only in unusual circumstances
do the messages need to convey further information, usually as an optimization to short circuit
extra queries.

This change set makes messages of this sort sort automatic. It:

 - Defines a convention for generating media types based on a resource and a lifecycle change
 - Defines a base `URIMessageSchema` that can be used (instead of defining custom schemas for every message)
 - Extends the schema registry to (optionally) be aware of these messages
 - Defines semantic sugar to easily construct such messages.

Example #1:

    # producing a message without defining a schema
    graph.sns_producer.produce(created(MyResource), uri="http://example.com", opaque_data=dict())

Example #2:

    # registering a message handler without defining a schema
    @handles(created(MyResource))
    def my_handler(message):
        pass

With these conventions, micro services (mostly) do not need to define message types for pubsub processing.

### DIFF
--- a/microcosm_pubsub/conventions.py
+++ b/microcosm_pubsub/conventions.py
@@ -1,0 +1,98 @@
+"""
+Convention-driven pubsub.
+
+"""
+from enum import Enum, unique
+
+from inflection import underscore
+from marshmallow import fields
+
+from microcosm_pubsub.codecs import PubSubMessageSchema
+
+
+@unique
+class LifecycleChange(Enum):
+    """
+    CRUD lifecycle changes that may be announced via pubsub.
+
+    Note that `Changed` is strongly discouraged; use immutable resources instead.
+
+    """
+    Changed = u"changed"
+    Created = u"created"
+    Deleted = u"deleted"
+
+    def matches(self, media_type):
+        return self.value in media_type.split(".")
+
+
+def make_media_type(resource, lifecycle_change=None, organization=None, public=False):
+    """
+    Generate a canonical media type for a message that announces a lifecycle change to a resource.
+
+    :param resource: the resource (name)
+    :param lifecycle_change: the change that happened to the resource
+    :param organization: the vendor organization to which this resource belongs
+    :param public: whether the resource is meant for public/external consumption
+
+    """
+    if lifecycle_change is None:
+        lifecycle_change = LifecycleChange.Created
+    if organization is None:
+        organization = "globality"
+
+    return "application/vnd.{}.{}.{}.{}".format(
+        # use a vendor specific media type
+        organization,
+        # differentiate messages sent over a pubsub channel from other messages
+        "pubsub" if public else "pubsub._",
+        # qualify the message with the kind of lifecycle change
+        lifecycle_change.value,
+        # specify the resource name
+        underscore(resource),
+    )
+
+
+class URIMessageSchema(PubSubMessageSchema):
+    """
+    Define a baseline message schema that points to (the URI of) a resource that experienced a lifecycle change.
+
+    By convention, pubsub messages should be *references* to something that happened within some other
+    source-of-truth (e.g. a CRUD microservice). Because there are inherent race conditions between publishing
+    a message and commit a persistent transaction, message consumers are expected to call back to source-of-truth
+    (e.g. via HTTP) and fetch the current resource value. In the event that message has not been committed yet,
+    the consumer can retry a few times before giving up (e.g. via SQS dead-lettering).
+
+    """
+    def __init__(self, media_type, **kwargs):
+        super(PubSubMessageSchema, self).__init__(**kwargs)
+        self.MEDIA_TYPE = media_type
+
+    uri = fields.String(required=True)
+
+    def deserialize_media_type(self, obj):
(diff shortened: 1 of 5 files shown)


microcosm_pubsub/conventions.py
```diff
@@ -0,0 +1,98 @@
...
+@unique
+class LifecycleChange(Enum):
+    """
+    CRUD lifecycle changes that may be announced via pubsub.
+
+    Note that `Changed` is strongly discouraged; use immutable resources instead.
+
+    """
+    Changed = u"changed"
+    Created = u"created"
```
jessemyers: I went with `Created` instead of `New`; this means that these messages will not interchange with the existing message schemas and we'll have to change both sides of every interaction.

I was torn about this; the migration to not defining schemas in every repo will be easier if we have backwards compatibility. However, these adjectives are a lot more consistent/clear with the rest of the system.

microcosm_pubsub/conventions.py
```diff
@@ -0,0 +1,98 @@
...
+    a message and commit a persistent transaction, message consumers are expected to call back to source-of-truth
+    (e.g. via HTTP) and fetch the current resource value. In the event that message has not been committed yet,
+    the consumer can retry a few times before giving up (e.g. via SQS dead-lettering).
+
+    """
+    def __init__(self, media_type, **kwargs):
+        super(PubSubMessageSchema, self).__init__(**kwargs)
+        self.MEDIA_TYPE = media_type
+
+    uri = fields.String(required=True)
```
jessemyers: I'm deliberately omitting any notion of `context` at this point. We do have use cases for handlers pre-filtering messages before fetching the URI, but I'm not sure they are as important as we think they are.

microcosm_pubsub/registry.py
```diff
@@ -90,8 +93,25 @@ def __getitem__(self, media_type):
         Create a codec or raise KeyError.
 
         """
-        schema_cls = self.mappings[media_type]
-        return PubSubMessageCodec(schema_cls(strict=self.strict))
+        try:
+            schema_cls = self.mappings[media_type]
+            if isinstance(schema_cls, string_types):
```
jessemyers: If we use `@handles(created("Foo"))`, we will register a string as the media type; in this case, we assume the URI schema.

microcosm_pubsub/registry.py
```diff
@@ -90,8 +93,25 @@ def __getitem__(self, media_type):
...
+                media_type = schema_cls
+                schema = URIMessageSchema(media_type)
+            else:
+                schema = schema_cls(strict=self.strict)
+        except KeyError:
+            if not self.auto_register:
+                raise
+
+            for lifecycle_change in LifecycleChange:
```
jessemyers: If we miss the registry and have auto registering enabled, we can generate schema instance on the fly.

microcosm_pubsub/conventions.py
```diff
@@ -0,0 +1,98 @@
...
+
+    return "application/vnd.{}.{}.{}.{}".format(
+        # use a vendor specific media type
+        organization,
+        # differentiate messages sent over a pubsub channel from other messages
+        "pubsub" if public else "pubsub._",
+        # qualify the message with the kind of lifecycle change
+        lifecycle_change.value,
+        # specify the resource name
+        underscore(resource),
```
wheremiah: Should this be the name of the resource?

jessemyers: At the moment, it's assumed to be the resource name. I was contemplating duplicating [this function](https://github.com/globality-corp/microcosm-flask/blob/develop/microcosm_flask/naming.py#L11) so we can pass in classes as well.

wheremiah:
LGTM

afallou:
LGTM

microcosm_pubsub/conventions.py
adamhadani: @jessemyers this is a pretty major use case - for the text highlighting for example right now we're intercepting all qna event pubsub messages and pre-filtering based on the 'tag' - otherwise we'd be materializing very many QnAEvents for no reason.. I think makes sense to have a notion of a 'header' for a pubsub message that can vary based on message type and contains information useful for pre-filtering - we can formalize that notion more carefully so is not as ad hoc as it was..

jessemyers: I agree that's a valid use case, but I think it's 10% (or less) use case. I'd like to start by converting the messages that don't have this need and see what's left. Some of the leftovers will be valid (like your example); some won't be.

jessemyers:
Please address naming vagueness and schema migrations.

microcosm_pubsub/conventions.py
adamhadani: 👍

PR #33, opened 2016-10-13
Add opaque data to SNS logging

jessemyers:


microcosm_pubsub/producer.py
```diff
@@ -84,8 +93,11 @@ def __init__(self, producer):
         self.producer = producer
         self.messages = []
 
-    def produce(self, media_type, dct=None, **kwargs):
-        message = self.producer.create_message(media_type, dct, **kwargs)
+    def produce(self, media_type, dct=None, opaque_data=None, **kwargs):
+        if opaque_data is None:
+            opaque_data = dict()
```
cdm-ium: No `opaque.as_dict` call here?

cdm-ium:
+1 one logical question, but I may not be reading closely enough.

jessemyers:
Good catch. I've fixed the logic.

PR #34, opened 2016-10-26
format log messages

sethisernhagen:
- use str.format() for log messages

### DIFF
--- a/microcosm_pubsub/handlers.py
+++ b/microcosm_pubsub/handlers.py
@@ -64,7 +64,9 @@ class URIHandler(object):
 
     def on_call(self, message, uri):
         self.logger.debug(
-            "Starting {handler}",
+            "Starting {handler}".format(
+                handler=self.name,
+            ),
             extra=dict(
                 handler=self.name,
                 uri=uri,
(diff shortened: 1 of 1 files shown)


microcosm_pubsub/handlers.py
```diff
@@ -64,7 +64,9 @@ def __call__(self, message):
 
     def on_call(self, message, uri):
         self.logger.debug(
-            "Starting {handler}",
+            "Starting {handler}".format(
```
cdm-ium: Using a string format here is redundant. The console logger will do the formatting for you and in loggly the format string will show up as is (so you can search by it) and the values in extra will be added to the Json.

cdm-ium:
The way the formatters in `microcosm_logging` are intended to work are:

- There is a console logger which will format values from extra into the string specified in the log message
- There is a json logger which will *not* format values from extra into the log message instead it puts the values from extra in the json that is logged and maintains the format string as is so you can search by it.

sethisernhagen:
@cdm-ium Thanks for the explanation!  This is what I'm seeing in diego logs right now -```json:
message: Handled {handler}
handler: Sendnewworkspaceemail
uri: https://juvenal.dev.globality.io:5403/api/v2/service_provider_match/92ab6066-d9d2-4688-9215-8d6e8dc9fcc0```

cdm-ium:
@sethisernhagen That's expected for what should show up in loggly. The logs on the box should have things actually formatted in.

sethisernhagen:
@cdm-ium Got it. Sorry for the confusion. Closing.

cdm-ium:
One bonus feature that we could add is having both the format string and the formatted string show up in two separate fields in loggly for reading pleasure as well as searchability.

jessemyers:
@cdm-ium please!

PR #35, opened 2016-12-13
fix type errors on SQS limits pulled from credstash

tomgoren:


jessemyers:
LGTM if you add my fix

microcosm_pubsub/consumer.py
```diff
@@ -85,8 +85,8 @@ def configure_sqs_consumer(graph):
 
     """
     sqs_queue_url = graph.config.sqs_consumer.sqs_queue_url
-    limit = graph.config.sqs_consumer.limit
-    wait_seconds = graph.config.sqs_consumer.wait_seconds
+    limit = int(graph.config.sqs_consumer.limit)
+    wait_seconds = int(graph.config.sqs_consumer.wait_seconds)
     visibility_timeout_seconds = graph.config.sqs_consumer.visibility_timeout_seconds
```
tomgoren: Thanks! As mentioned offline, the default value is already set above to `None`.

PR #36, opened 2016-12-13
Release 0.14.0: Set type for values returned from config object

tomgoren:
We ran into issues related to DynamoDB storing everything as unicode. When
we pull the config from credstash we encountered type mismatches.

This should fix it for the SQS limits and timeouts.


jessemyers:
LGTM

PR #40, opened 2017-01-23
Pinned marshmallow to latest known working version

tomgoren:
https://github.com/marshmallow-code/marshmallow/pull/572#issuecomment-274433782
caused builds to break. Until
https://github.com/marshmallow-code/marshmallow/pull/575 is figured out,
we will pin to `2.11.1` which is known to work.


jessemyers:
LGTM

PR #45, opened 2017-02-08
Add a SkipMessage exception in the base pubsub path.

jessemyers:


microcosm_pubsub/dispatcher.py
```diff
@@ -74,6 +83,9 @@ def handle_message(self, media_type, message):
                     parent=sqs_message_handler,
                 )
                 return handler_with_context(message)
+            except SkipMessage as skipped:
+                self.logger.debug("Skipping message for reason: {}".format(skipped.message))
```
cdm-ium: I might have moved this down to the block below to use the handler logger so you know which handler decided to skip the message. Again, this is a debug log so who knows when anyone would see it, but it seems useful at least for consistency in the logging.

jessemyers: See: https://github.com/globality-corp/microcosm-pubsub/pull/46

PR #46, opened 2017-02-08
Hotfix/0.24.2

jessemyers:
Spread out error handling

### DIFF
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -83,9 +83,11 @@ class SQSMessageDispatcher(object):
                 )
                 return handler_with_context(message)
             except SkipMessage as skipped:
+                extra = self.sqs_message_context(message)
+                extra.update(skipped.extra)
                 logger.info(
                     "Skipping message for reason: {}".format(str(skipped)),
-                    extra=skipped.extra,
+                    extra=extra,
                 )
                 return False
             except Nack:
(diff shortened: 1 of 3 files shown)


microcosm_pubsub/dispatcher.py
```diff
@@ -76,34 +83,27 @@ def handle_message(self, media_type, message):
                 )
                 return handler_with_context(message)
             except SkipMessage as skipped:
-                self.logger.info(
+                logger.info(
                     "Skipping message for reason: {}".format(str(skipped)),
                     extra=skipped.extra,
                 )
                 return False
+            except Nack:
```
jessemyers: We should not use `instanceof` on errors.

microcosm_pubsub/dispatcher.py
```diff
@@ -76,34 +83,27 @@ def handle_message(self, media_type, message):
                 )
                 return handler_with_context(message)
             except SkipMessage as skipped:
-                self.logger.info(
+                logger.info(
```
jessemyers: It's unclear if I should merge the message context with the extra data here.

cdm-ium:
+1 nice thanks.

microcosm_pubsub/dispatcher.py
cdm-ium: I would probably merge unless we add a base pubsub handler that automatically merges the context into extra. No way everyone raising the exception will remember to put the message context in there on their own.

PR #48, opened 2017-03-29
Allow convention-oriented topic mappings to be set via environment variables

jessemyers:
We increasingly implement configuration via environment variables, but we our
pubsub media type conventions tend to use naming schemas that can't be expressed
using environment variables. In particular, using previous logic, we'd want to
generate a configuration dictionar with something like:

    {
      "sns_topic_arns": {
        "mappings": {
          "application/vnd.globality.pubsub._.created.foo_bar.baz": "topic-name",
        }
      }
    }

While the microcosm environment variable loader can handle nested dictionaries
(via double underscores), we cannot express '/' or '.' because they aren't legal
characters.

We solve this problem in two ways:

 1. We add support for the convention-driven media types using alternative keys (avoiding '/')
 2. We (re)collapse double underscores within these convention keys (avoiding '.')

That is, we implement the above mapping using:

    {
      "sns_topic_arns": {
        "created": {
          "foo_bar": {
            "baz": "topic-name",
          },
        }
      }
    }

### DIFF
--- a/microcosm_pubsub/producer.py
+++ b/microcosm_pubsub/producer.py
@@ -3,6 +3,7 @@ Message producer.
 
 """
 from collections import defaultdict
+from six import string_types
 
 from boto3 import client
 from microcosm.api import defaults
@@ -10,6 +11,7 @@ from microcosm.errors import NotBoundError
 from microcosm_logging.decorators import logger
 from microcosm_logging.timing import elapsed_time
 
+from microcosm_pubsub.conventions import LifecycleChange, make_media_type
 from microcosm_pubsub.errors import TopicNotDefinedError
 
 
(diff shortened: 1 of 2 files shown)


KensoDev:
LGTM

microcosm_pubsub/producer.py
```diff
@@ -109,6 +111,33 @@ def __exit__(self, type, value, traceback):
...
+        else:
+            yield separator.join([prefix, key]), value
+
+
+def iter_topic_mappings(dct):
+    for key, value in dct.items():
+        if isinstance(value, string_types):
+            yield key, value
+        else:
+            for key, value in collapse_dict(value, key):
```
pelegdvir: Rename (key, value) names (in the left side) - you already used the key, value names

jessemyers: You can't return from inside a generator; you have to yield.

I fixed the naming.

microcosm_pubsub/tests/test_producer.py
```diff
@@ -115,6 +118,42 @@ def loader(metadata):
...
+                ),
+                baz=dict(
+                    foo=dict(
+                        bar="foo",
+                    ),
+                ),
+            )
+        )
+    )
+    assert_that(result, is_(equal_to({
```
pelegdvir: Can you expand this test - add  to the end something like `"bar.foo2": "baz2"`

PR #49, opened 2017-04-14
Some clean up in the pubsub registry code.

jessemyers:
There's some nastiness in the registry logic right primarily to handle
legacy and test behaviors. As a first step, remove the registry base
class and clean up the registry interfaces.

### DIFF
--- a/microcosm_pubsub/daemon.py
+++ b/microcosm_pubsub/daemon.py
@@ -16,7 +16,7 @@ class ConsumerDaemon(Daemon):
 
     def run_state_machine(self):
         for media_type in self.graph.sqs_message_handler_registry.keys():
-            handler = self.graph.sqs_message_handler_registry[media_type]
+            handler = self.graph.sqs_message_handler_registry.find(media_type)
             self.graph.logger.info("Handling: {} with handler: {}".format(
                 media_type,
                 handler.__class__.__name__,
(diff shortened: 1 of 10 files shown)


microcosm_pubsub/daemon.py
```diff
@@ -16,7 +16,7 @@ def make_arg_parser(self):
 
     def run_state_machine(self):
         for media_type in self.graph.sqs_message_handler_registry.keys():
-            handler = self.graph.sqs_message_handler_registry[media_type]
+            handler = self.graph.sqs_message_handler_registry.find(media_type)
```
jessemyers: Use `find()` instead of `__getitem__`.

microcosm_pubsub/registry.py
```diff
@@ -16,21 +16,17 @@ class AlreadyRegisteredError(Exception):
...
+    Keeps track of available message schemas.
 
     """
     def __init__(self, graph):
-        """
-        Create registry.
-
-        """
-        self.mappings = dict()
+        self._mappings = dict()
```
jessemyers: Ensure that the member variable is internal.

microcosm_pubsub/tests/fixtures.py
```diff
@@ -23,8 +25,16 @@ def name(self):
         return "example"
 
 
+class Foo(object):
```
jessemyers: Unify fixtures.

netama:
LGTM

PR #50, opened 2017-04-17
Simplify and clarify pubsub registry interactions

jessemyers:
Main changes:
 -  Clearly differentiate convention-driven and class-driven schema registration
 -  Handle cases with multiple daemons registering different handlers for the same media type
 -  Make fixtures more obvious

### DIFF
--- a/microcosm_pubsub/daemon.py
+++ b/microcosm_pubsub/daemon.py
@@ -9,18 +9,30 @@ from microcosm_daemon.daemon import Daemon
 
 class ConsumerDaemon(Daemon):
 
+    def __init__(self):
+        super(ConsumerDaemon, self).__init__()
+        self.bound_handlers = None
+
     def make_arg_parser(self):
         parser = super(ConsumerDaemon, self).make_arg_parser()
         parser.add_argument("--sqs-queue-url")
         return parser
 
+    def create_object_graph_components(self, graph):
+        super(ConsumerDaemon, self).create_object_graph_components(graph)
+        self.bound_handlers = graph.sqs_message_handler_registry.compute_bound_handlers(
+            self.components,
+        )
+
     def run_state_machine(self):
-        for media_type in self.graph.sqs_message_handler_registry.keys():
-            handler = self.graph.sqs_message_handler_registry.find(media_type)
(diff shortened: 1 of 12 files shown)


microcosm_pubsub/daemon.py
```diff
@@ -9,18 +9,30 @@
...
+        self.bound_handlers = None
+
     def make_arg_parser(self):
         parser = super(ConsumerDaemon, self).make_arg_parser()
         parser.add_argument("--sqs-queue-url")
         return parser
 
+    def create_object_graph_components(self, graph):
+        super(ConsumerDaemon, self).create_object_graph_components(graph)
+        self.bound_handlers = graph.sqs_message_handler_registry.compute_bound_handlers(
```
jessemyers: It's possible to have multiple daemons in the same code base; in this case, the handler registry could have more than one entry. We solve this by determining which handlers are part of the daemon's object graph and restricting access to just those handlers.

microcosm_pubsub/tests/fixtures.py
```diff
@@ -4,67 +4,64 @@
...
-        return "example"
-
-
-class Foo(object):
-    pass
-
-
 @schema
-class FooSchema(PubSubMessageSchema):
+class DerivedSchema(PubSubMessageSchema):
```
jessemyers: I tried to get rid of `foo` and `bar` and be more explicit with my naming. This avoids some spurious test cases where we have duplicate names.

microcosm_pubsub/daemon.py
```diff
@@ -9,18 +9,30 @@
...
+            self.components,
+        )
+
     def run_state_machine(self):
-        for media_type in self.graph.sqs_message_handler_registry.keys():
-            handler = self.graph.sqs_message_handler_registry.find(media_type)
+        self.compute_bound_handlers()
```
netama: where is this defined?

netama:
LGTM with one question

microcosm_pubsub/daemon.py
jessemyers: oops, that's from before a refactor

PR #52, opened 2017-04-20
Simplify testing real daemons with fake queues

jessemyers:
This allows us to run:

    some-consume-daemon --sqs-queue-url test

And exercise behavior not tied to SQS consumption.

### DIFF
--- a/microcosm_pubsub/consumer.py
+++ b/microcosm_pubsub/consumer.py
@@ -95,6 +95,9 @@ def configure_sqs_consumer(graph):
     if graph.metadata.testing:
         from mock import MagicMock
         sqs_client = MagicMock()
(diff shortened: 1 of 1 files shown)


microcosm_pubsub/consumer.py
```diff
@@ -95,6 +95,9 @@ def configure_sqs_consumer(graph):
     if graph.metadata.testing:
         from mock import MagicMock
         sqs_client = MagicMock()
+    elif sqs_queue_url == "test":
```
jessemyers: I needed this to test standby behavior with a real consumer daemon and not spew errors.

It would be slightly better to use something other than a `MagicMock` since we may not have that imported outside of local development.

KensoDev:
LGTM

PR #53, opened 2017-05-02
Use identity-based convention for deleted lifecycle changes

jessemyers:
When a resource is deleted, it doesn't make sense to depend on a URI-based
message since the underlying object will usually not be accessible at its
URI any more. Use an identity-based schema instead.

### DIFF
--- a/microcosm_pubsub/registry.py
+++ b/microcosm_pubsub/registry.py
@@ -10,7 +10,7 @@ from microcosm.api import defaults
 from microcosm_logging.decorators import logger
 
 from microcosm_pubsub.codecs import PubSubMessageCodec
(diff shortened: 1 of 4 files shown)


microcosm_pubsub/registry.py
```diff
@@ -72,7 +72,10 @@ def find(self, media_type):
             schema = schema_cls(strict=self.strict)
         except KeyError:
             # use convention otherwise
-            schema = URIMessageSchema(media_type)
+            if LifecycleChange.Deleted.value in media_type.split("."):
```
jessemyers: This could be prettier, but I'm ok with it for now.

pelegdvir:
LGTM

PR #54, opened 2017-05-22
Create decorator to allow for automatic message deferment

GabeLambda:
This decorator is intended to be used at the controller configuration
level, with the idea being that any messages produced within the context of a
controller function will be produced after it has completed.

This will allow these messages to be published after the resource
they're referencing has been persisted, guaranteeing that it will be available
when fetched by other message handlers.

Note that this is not a full solution; it's not uncommon for components
to call other components directly, and this decorator will not change those
references in place. As is, it will only apply the messages produced within a
function by the specified component.

### DIFF
--- a/microcosm_pubsub/tests/test_producer.py
+++ b/microcosm_pubsub/tests/test_producer.py
@@ -19,7 +19,7 @@ import microcosm.opaque  # noqa
 
 from microcosm_pubsub.conventions import created
 from microcosm_pubsub.errors import TopicNotDefinedError
-from microcosm_pubsub.producer import DeferredProducer, iter_topic_mappings
+from microcosm_pubsub.producer import deferred, DeferredProducer, iter_topic_mappings
 from microcosm_pubsub.tests.fixtures import DerivedSchema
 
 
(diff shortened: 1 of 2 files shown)


microcosm_pubsub/tests/test_producer.py
```diff
@@ -165,3 +165,33 @@ def loader(metadata):
...
+            assert isinstance(self.sns_producer, DeferredProducer)
+            self.sns_producer.produce(DerivedSchema.MEDIA_TYPE, data="data")
+
+    graph = create_object_graph("example", testing=True, loader=loader)
+    foo = Foo(graph)
+
+    func = deferred(foo)(foo.bar)
+    func()
+
+    assert_that(graph.sns_producer.sns_client.publish.call_count, is_(equal_to(1)))
```
wheremiah: This tests confirms that the function is called, but not that the call is deferred.  Is it possible to add an assert inside `bar()` to show that `produce ()` is is not called right away?

GabeLambda: It's inferred. `bar` above makes sure that the producer it's getting has been swapped out for a `DeferredProducer`, making the entire block deferred. I can make it more explicit.

jessemyers:
LGTM

Let me know when you've merged and I'll cut a new release.

microcosm_pubsub/tests/test_producer.py
wheremiah: Ah, got it, thx.

PR #55, opened 2017-05-23
Generalize SNS and SQS support for LocalStack

jessemyers:
-  Add `endpoint_url` to boto3 client creation
 -  Generalize envelope parsing to support LocalStack encoding

### DIFF
--- a/microcosm_pubsub/envelope.py
+++ b/microcosm_pubsub/envelope.py
@@ -9,6 +9,7 @@ process this envelope, depending on the degree of validation and metadata desire
 from abc import ABCMeta, abstractmethod
 from json import loads
 from hashlib import md5
+from six import add_metaclass
 
 from microcosm.api import defaults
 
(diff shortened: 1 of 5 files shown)


microcosm_pubsub/envelope.py
```diff
@@ -9,20 +9,106 @@
...
+from six import add_metaclass
 
 from microcosm.api import defaults
 
 from microcosm_pubsub.codecs import MediaTypeSchema, PubSubMessageCodec
 from microcosm_pubsub.message import SQSMessage
 
 
-class SQSEnvelope(object):
+@add_metaclass(ABCMeta)
```
jessemyers: Big picture here:
 - We have two parsing steps, one that pulls out "our" parts of the message from SQS and one that translates this message into a media type and dictionary.
 - When working with `LocalStack` the first of these functions is abnormal.
 - We already had some support for generalizing the second of these parse functions
 - The changes here rewrite the SQS envelope parse into two abstract classes so we can mix and match

microcosm_pubsub/tests/test_envelope.py
```diff
@@ -0,0 +1,102 @@
+"""
+SQS envelope tests.
```
jessemyers: New: tests for the various envelope formats

microcosm_pubsub/tests/test_envelope.py
```diff
@@ -0,0 +1,102 @@
...
+    receipt_handle = "receipt_handle"
+    envelope = CodecSQSEnvelope(graph)
+
+    media_type = created("foo")
+    uri = "http://foo/id"
+
+    sqs_message = envelope.parse_raw_message(consumer, dict(
+        MessageId=message_id,
+        ReceiptHandle=receipt_handle,
+        Body=dumps(dict(
```
jessemyers: This is what we normally get: a double encoded doc; there's one layer of JSON encoding from SNS to SQS and one from our codec

microcosm_pubsub/tests/test_envelope.py
```diff
@@ -0,0 +1,102 @@
...
+    receipt_handle = "receipt_handle"
+    envelope = LocalStackSQSEnvelope(graph)
+
+    media_type = created("foo")
+    uri = "http://foo/id"
+
+    sqs_message = envelope.parse_raw_message(consumer, dict(
+        MessageId=message_id,
+        ReceiptHandle=receipt_handle,
+        Body=dumps(dict(
```
jessemyers: LocalStack only has one level of encoding because it doesn't faithfully preserve SNS subscription semantics

afallou:
+1

PR #56, opened 2017-05-23
Add elapsed time to `handle` processing

GabeLambda:
This data is can be traced, but gathering the overall time of execution is difficult.

### DIFF
--- a/microcosm_pubsub/handlers.py
+++ b/microcosm_pubsub/handlers.py
@@ -7,6 +7,7 @@ from inflection import titleize
 
 from requests import codes, get
 
+from microcosm_logging.timing import elapsed_time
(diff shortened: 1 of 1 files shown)


microcosm_pubsub/handlers.py
```diff
@@ -55,11 +56,18 @@ def __call__(self, message):
 
         resource = self.get_resource(message, uri)
 
-        if self.handle(message, uri, resource):
-            self.on_handle(message, uri, resource)
+        extra = dict(
```
jessemyers: I'm ok with this, but I think the more correct approach is to inject into the graph's current opaque dictionary.

@cdm-ium can help you with that

jessemyers:
LGTM but see note

PR #57, opened 2017-06-13
Adds loads on content in raw envelope

wheremiah:
@jessemyers  - Is this correct or should be create our own strategy

### DIFF
--- a/microcosm_pubsub/envelope.py
+++ b/microcosm_pubsub/envelope.py
@@ -66,7 +66,7 @@ class RawMediaTypeAndContentParser(MediaTypeAndContentParser):
     def parse_media_type_and_content(self, message):
         content = message
         media_type = "application/json"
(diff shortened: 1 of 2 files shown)


jessemyers:
It seems correct, but unit tests are failing?

sethisernhagen:
I think we need to do dumps() here https://github.com/globality-corp/microcosm-pubsub/blob/develop/microcosm_pubsub/tests/test_envelope.py#L32

```
    sqs_message = envelope.parse_raw_message(consumer, dict(
        MessageId=message_id,
        ReceiptHandle=receipt_handle,
        Body=dumps(dict(
            foo="bar",
        )),
    ))
```

wheremiah:
@sethisernhagen I think you are right, thx!

jessemyers:
LGTM

PR #58, opened 2017-07-14
Extract approximate receive count on SQS receive

jessemyers:
We can use this count for backoff purposes.

It's documented as being approximate, but it seems to be accurate empirically. 

I also added (back) a CLI for testing pubsub using standard conventions:

```
# write a message; note that the SNS id is different from the SQS id
> pubsub write --topic-arn $TOPIC --uri http://example.com --media-type example
Wrote SNS message: 59d07afd-0e38-5f2b-8708-2b17e094ce1d

# read the same message back; the parenthetical includes the read count
> pubsub read --queue-url $QUEUE --nack
Read SQS message: 67859b2d-94e3-43c5-8d11-a863796901f3 (1)

> pubsub read --queue-url $QUEUE --nack
Read SQS message: 67859b2d-94e3-43c5-8d11-a863796901f3 (2)

> pubsub read --queue-url $QUEUE --nack
Read SQS message: 67859b2d-94e3-43c5-8d11-a863796901f3 (3)

> pubsub read --queue-url $QUEUE --nack
Read SQS message: 67859b2d-94e3-43c5-8d11-a863796901f3 (4)
```

### DIFF
--- a/microcosm_pubsub/consumer.py
+++ b/microcosm_pubsub/consumer.py
@@ -36,8 +36,11 @@ class SQSConsumer(object):
         return [
             self.sqs_envelope.parse_raw_message(self, raw_message)
             for raw_message in self.sqs_client.receive_message(
-                QueueUrl=self.sqs_queue_url,
+                AttributeNames=[
+                    "ApproximateReceiveCount",
+                ],
                 MaxNumberOfMessages=self.limit,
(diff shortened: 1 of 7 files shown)


netama:
:+1:

wheremiah:
❤️

PR #59, opened 2017-07-20
Add support for configurable SQS consumer backoff policies

jessemyers:
Implements the current (naive) backoff policy and an experimental exponential backoff.

### DIFF
--- a/microcosm_pubsub/backoff.py
+++ b/microcosm_pubsub/backoff.py
@@ -1,0 +1,66 @@
+"""
+Consumer backoff policy
+
+"""
+from abc import ABCMeta, abstractmethod
+from random import randint
+from six import add_metaclass
+
+
+# we cannot go over 12 hours
+MAX_BACKOFF_TIMEOUT = 60 * 60 * 12
+
+
+@add_metaclass(ABCMeta)
+class BackoffPolicy(object):
+
+    def __init__(self, visibility_timeout_seconds=None):
+        self.visibility_timeout_seconds = visibility_timeout_seconds
+
+    @abstractmethod
+    def compute_backoff_timeout(self, message, message_timeout):
+        pass
+
+    @classmethod
+    def choose_backoff_policy(cls, name):
+        for subclass in cls.__subclasses__():
+            if subclass.__name__ == name:
+                return subclass
+        raise Exception("No backoff policy configured with class name: {}".format(name))
+
+
+class NaiveBackoffPolicy(BackoffPolicy):
+    """
+    Naive backoff policy.
+
+    Uses a fixed timeout from the message or system default.
+
+    """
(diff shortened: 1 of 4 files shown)


microcosm_pubsub/backoff.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @classmethod
+    def choose_backoff_policy(cls, name):
+        for subclass in cls.__subclasses__():
+            if subclass.__name__ == name:
+                return subclass
+        raise Exception("No backoff policy configured with class name: {}".format(name))
+
+
+class NaiveBackoffPolicy(BackoffPolicy):
```
jessemyers: This implements our current behavior. It either takes the default backoff or the message specific one. If neither are set, no backoff is applied and the queue's default will be used (30s?).

microcosm_pubsub/backoff.py
```diff
@@ -0,0 +1,66 @@
...
+    Uses a fixed timeout from the message or system default.
+
+    """
+    def compute_backoff_timeout(self, message, message_timeout):
+        backoff_timeout = message_timeout or self.visibility_timeout_seconds
+        # we can only set integer timeouts
+        return int(backoff_timeout) if backoff_timeout is not None else backoff_timeout
+
+
+class ExponentialBackoffPolicy(BackoffPolicy):
```
jessemyers: This policy does (almost) textbook exponential backoff except:
 - It's always integral (SQS limitation)
 - It never returns `0` (seems pointless, but I could be wrong)
 - It avoids returns more than 12 hours (SQS limitation)

microcosm_pubsub/consumer.py
```diff
@@ -65,20 +67,23 @@ def nack(self, message, visibility_timeout_seconds=None):
...
         """
-        timeout = visibility_timeout_seconds or self.visibility_timeout_seconds
-        if timeout is not None:
-            self.sqs_client.change_message_visibility(
-                QueueUrl=self.sqs_queue_url,
-                ReceiptHandle=message.receipt_handle,
-                # we can only set integer timeouts
-                VisibilityTimeout=int(timeout),
-            )
+        timeout = self.backoff_policy.compute_backoff_timeout(message, visibility_timeout_seconds)
```
jessemyers: Delegate timeout calculation to backoff policy

microcosm_pubsub/consumer.py
```diff
@@ -116,11 +121,19 @@ def configure_sqs_consumer(graph):
             region_name=region_name,
         )
 
+    backoff_policy_class = BackoffPolicy.choose_backoff_policy(
```
jessemyers: Choose the policy here.

afallou:
+1

PR #61, opened 2017-08-02
Save topic arn in message

jessemyers:


microcosm_pubsub/backoff.py
```diff
@@ -1,5 +1,5 @@
 """
-Consumer backoff policy
+oConsumer backoff policy
```
afallou: Typo?

wheremiah: typo

microcosm_pubsub/message.py
```diff
@@ -16,12 +16,14 @@ def __init__(self,
                  media_type,
                  message_id,
                  receipt_handle,
+                 topic_arn=None,
```
wheremiah: I think you also need to add this to the `parse_raw_message()` in microcosm_pubsub/envelope.py

PR #62, opened 2017-08-08
Allow daemons to run from a file URL and playback data files.

jessemyers:
This PR includes two new features:

 1. If the `--sqs-queue-url` points to a file, construct a simulated SQS client
    that reads JSON lines from a file until done.

 2. If the `NaiveSQSEnvelope` is used, allow a flat dictionary to be used for inputs
    (e.g. just `mediaType` and `uri`); this may be specified from the command line.

The net effect is that you can create `/path/to/file` with:

```
   {"mediaType": "some-media-type", "uri": "http://some-uri"}
   {"mediaType": "some-media-type", "uri": "http://some-uri"}
   {"mediaType": "some-media-type", "uri": "http://some-uri"}
   {"mediaType": "some-media-type", "uri": "http://some-uri"}
```

And run:

    my-consumer-daemon --debug --sqs-queue-url /path/to/file --envelope NaiveSQSEnvelope

And replay all of the file's messages.

### DIFF
--- a/microcosm_pubsub/envelope.py
+++ b/microcosm_pubsub/envelope.py
@@ -10,6 +10,7 @@ from abc import ABCMeta, abstractmethod
 from json import loads
 from hashlib import md5
 from six import add_metaclass
+from uuid import uuid4
 
 from microcosm.api import defaults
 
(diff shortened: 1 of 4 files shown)


microcosm_pubsub/envelope.py
```diff
@@ -69,6 +70,18 @@ def parse_media_type_and_content(self, message):
         return media_type, loads(content)
 
 
+class NaiveMediaTypeAndContentParser(MediaTypeAndContentParser):
```
guillaume-miara: Why do we need this new parser?

jessemyers: It's the difference between having your input file be:

```
{"mediaType": ..., "uri": ...}
```

And:
```
{"Body":"{\"mediaType\": ..., \"uri\": ...}"}
```

Or something equally awful.

It also lets us default some of the SQS required parameters (message id, receipt handle) that are irrelevant in this scenario.

PR #63, opened 2017-08-25
Cast ApproximateRecieveCount to int

GabeLambda:
Messages are strings, not typed values, and need to be casted.

This would otherwise raise an exception during the nack phase, and thus
fallback to being retried on the default visibility timeout.

I'm still tad confused as to where the exception was being swallowed; it made tracking down this bug a little harder.

### DIFF
--- a/microcosm_pubsub/envelope.py
+++ b/microcosm_pubsub/envelope.py
@@ -133,7 +133,7 @@ class SQSEnvelope(MessageBodyParser, MediaTypeAndContentParser):
         message_id = self.parse_message_id(raw_message)
         receipt_handle = self.parse_receipt_handle(raw_message)
         attributes = raw_message.get("Attributes", {})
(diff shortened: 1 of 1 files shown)


microcosm_pubsub/envelope.py
```diff
@@ -133,7 +133,7 @@ def parse_raw_message(self, consumer, raw_message):
         message_id = self.parse_message_id(raw_message)
         receipt_handle = self.parse_receipt_handle(raw_message)
         attributes = raw_message.get("Attributes", {})
-        approximate_receive_count = attributes.get("ApproximateReceiveCount")
+        approximate_receive_count = int(attributes.get("ApproximateReceiveCount", 1))
```
GabeLambda: I'm not sure if this is wholly correct here, but it will allow a fallback to a 1 second timeout if not provided.

PR #64, opened 2017-09-19
Default to publishing messages, even in debug

jessemyers:
We used to automatically suppress pubsub messages when running in debug (e.g. dev servers).
Now that `LocalStack` is good enough to run as a local alternative to AWS, this makes less
sense and actively breaks some forms of local integration testing.

### DIFF
--- a/microcosm_pubsub/producer.py
+++ b/microcosm_pubsub/producer.py
@@ -195,7 +195,7 @@ def configure_sns_topic_arns(graph):
     region_name=None,
     endpoint_url=None,
     mock_sns=True,
-    skip=None,
+    skip="False",
 )
 def configure_sns_producer(graph):
     """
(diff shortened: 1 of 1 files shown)


microcosm_pubsub/producer.py
```diff
@@ -230,13 +230,8 @@ def configure_sns_producer(graph):
     except NotBoundError:
         opaque = None
 
-    if graph.config.sns_producer.skip is None:
```
jessemyers: Note that with this change, running `runserver` or the equivalent and making HTTP requests with pubsub side effects *WILL* error out if there's no where to publish to. This means that running with `LocalStack` is more or less mandatory.

I'm open to alternatives.

netama: It would be nice not to have to worry about this when debugging locally. can we keep the default and change the ansible default instead?

jessemyers: We can change the globe config defaults probably.

PR #65, opened 2017-12-14
GLOB-15482 Add deferred batch producer and handler.

cdm-ium:
Why? For complicated routes, the deferred producer is insufficient for
avoiding NACKs caused by messages being published before a transaction
is committed. This adds a mechanism for batching messages that would be
published by a routed and republishing them from a daemon handler
*after* the transaction that wants to publish them is commited.

### DIFF
--- a/microcosm_pubsub/handlers.py
+++ b/microcosm_pubsub/handlers.py
@@ -131,7 +131,7 @@ class URIHandler(object):
 @binding("publish_message_batch")
 @handles(created("BatchMessage"))
 @logger
-class PublishBatchMessage:
+class PublishBatchMessage(object):
 
     def __init__(self, graph):
         self.sns_producer = graph.sns_producer
--- a/microcosm_pubsub/handlers.py
+++ b/microcosm_pubsub/handlers.py
@@ -7,6 +7,10 @@ from inflection import titleize
 
 from requests import codes, get
 
+from microcosm.api import binding
+from microcosm_logging.decorators import logger
+from microcosm_pubsub.conventions import created
+from microcosm_pubsub.decorators import handles
 from microcosm_pubsub.errors import Nack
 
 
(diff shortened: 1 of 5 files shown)


microcosm_pubsub/handlers.py
```diff
@@ -122,3 +126,22 @@ def get_resource(self, message, uri):
 
     def handle(self, message, uri, resource):
         return True
+
+
+@binding("publish_message_batch")
+@handles(created("BatchMessage"))
+@logger
+class PublishBatchMessage(object):
```
cdm-ium: I'm guessing we want `object` here for Python2 compatibility?

microcosm_pubsub/handlers.py
```diff
@@ -122,3 +126,22 @@ def get_resource(self, message, uri):
 
     def handle(self, message, uri, resource):
         return True
+
+
+@binding("publish_message_batch")
```
cdm-ium: My thinking is that projects that use the deferred producer can simply require the `publish_message_batch` handler in the appropriate daemon.

microcosm_pubsub/handlers.py
jessemyers: Until we decide to stop supporting Python2.

microcosm_pubsub/tests/test_producer.py
```diff
@@ -196,3 +202,62 @@ def bar(self):
...
+
+def test_deferred_batch_production():
+    """
+    Deferred production waits until the end of a block and publishes all
+    messages in one MessageBatchSchema
+
+    """
+    def loader(metadata):
+        return dict(
+            sns_topic_arns=dict(
```
jessemyers: It would be really nice to have the validation that requires the sns topic arn not blow up in test mode.

microcosm_pubsub/producer.py
```diff
@@ -116,6 +116,27 @@ def __exit__(self, type, value, traceback):
...
+                media_type=media_type,
+                message=message,
+                topic_arn=topic_arn,
+                opaque_data=opaque_data,
+            )
+            for media_type, message, topic_arn, opaque_data in self.messages
+        ]
+
+        self.producer.produce(
+            MessageBatchSchema.MEDIA_TYPE,
```
cdm-ium: One thing that occurs to me is that we might *not* want to allow publishing on the default topic for these messages or then *every* daemon (i.e. for different projects) with the batch message publisher will receive these messages and try to republish them. Thoughts on explicitly not letting this happen vs being careful and making sure we configure a different (service-specific) topic for these messages?

cdm-ium:
I updated the topic configuration to explicitly not use the defautl for the batch media type and added a test. @jessemyers

PR #66, opened 2017-12-15
GLOB-15482 Log elapsed time with daemon handler execution.

cdm-ium:
Why? It's useful to know how long your handlers take when trying to tune
flows that can involve multiple handlers each making multiple http
requests.

### DIFF
--- a/microcosm_pubsub/handlers.py
+++ b/microcosm_pubsub/handlers.py
@@ -82,7 +82,7 @@ class URIHandler(object):
         )
 
     def on_handle(self, message, uri, resource):
-        self.logger.info(
+        self.logger.debug(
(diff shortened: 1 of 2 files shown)


microcosm_pubsub/handlers.py
```diff
@@ -82,7 +82,7 @@ def on_skip(self, message, uri, reason):
         )
 
     def on_handle(self, message, uri, resource):
-        self.logger.info(
+        self.logger.debug(
```
cdm-ium: I'm basically moving this log a layer up so we can track elapsed time of the whole handler. I'm leaving the log because having `on_handle` in the `URIHandler` api is probably useful for other things.

PR #67, opened 2017-12-19
Rework batch messages so they are added to the registry.

cdm-ium:
Why? If they are not in the registry the message batch handler will not be
able to properly parse these handlers as it will fall back to use the
URIPubSubMessage Codec and ommit fields.

### DIFF
--- a/microcosm_pubsub/batch.py
+++ b/microcosm_pubsub/batch.py
@@ -1,0 +1,31 @@
+from marshmallow import fields, Schema
+
+from microcosm_pubsub.conventions import created
+from microcosm_pubsub.codecs import PubSubMessageSchema
+from microcosm_pubsub.decorators import schema
+
+
+class BatchedMessageSchema(Schema):
+    """
+    A wrapper for a single message that is to be published in a
+    MessageBatchSchema.
+
+    """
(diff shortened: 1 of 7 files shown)


microcosm_pubsub/batch.py
```diff
@@ -0,0 +1,31 @@
+from marshmallow import fields, Schema
```
cdm-ium: I had to move these to a separate module or I created a cyclic dependency between the `conventions` and the registry because we do something somewhat weird for the URIPubsubMessage. We don't wire it into the registry and instead we hardcode it as the default in the code itself.

microcosm_pubsub/registry.py
```diff
@@ -56,7 +56,8 @@ def register(self, media_type, value):
 
     def find(self, media_type):
         """
-        Create a codec or raise KeyError.
+        Create a codec or raise KeyError. If autoregistration is enabled, falls
```
cdm-ium: Not sure if this is good enough, but this was the magic that got me stuck for a while. In practice most of our code uses autoregistration to implicitly treat things with the URIMessageSchema so we don't have many examples of explicitly wiring messages into the schema in our code.

PR #70, opened 2018-02-23
Guard against repeated media type handlers

GabeLambda:
Report attempts to register multiple handlers for a for given media type.

Daemons will otherwise fail silently and pick only a single handler on
initialization.

We could ostensibly handle the same pubsub message multiple times within the same daemon, but it's probably preferred not to.

### DIFF
--- a/microcosm_pubsub/registry.py
+++ b/microcosm_pubsub/registry.py
@@ -115,6 +115,21 @@ class SQSMessageHandlerRegistry(object):
             if handler in bound_components or handler in bound_component_types
         }
 
+        media_type_handlers = {}
+        for media_type, handler in self.iter_handlers():
+            if handler in bound_components or handler in bound_component_types:
+                if media_type in media_type_handlers:
+                    raise AlreadyRegisteredError(
+                        "Handler {} already registered for media type: {}".format(
+                            handler.__name__,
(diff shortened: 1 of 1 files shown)


microcosm_pubsub/registry.py
```diff
@@ -115,6 +115,21 @@ def compute_bound_handlers(self, bindings):
             if handler in bound_components or handler in bound_component_types
         }
 
+        media_type_handlers = {}
```
netama: this is after a `return` statement

netama: worth adding a unittest?

cdm-ium: Move the if stream to a helper function and do a `dict` comprehension?

GabeLambda: Probably misunderstanding, not sure what you mean. This is before the `return`.

GabeLambda: Sure.

GabeLambda:
Tests are failing because we still technically haven't removed 2.7 support, but microcosm was updated to use 3.0+ syntax.

GabeLambda:
Also, will add a unit test, it's just slightly complicated with regards to graph / daemon setup.

PR #71, opened 2018-02-26
Guard against multiple handlers registered for the same daemon

GabeLambda:
Previously, we would silently ignore all but the last handler registered
for a media type for a given daemon. This makes that case an explicit
error during initialization.

### DIFF
--- a/microcosm_pubsub/registry.py
+++ b/microcosm_pubsub/registry.py
@@ -109,11 +109,19 @@ class SQSMessageHandlerRegistry:
             type(bound_component)
             for bound_component in bound_components
         ]
-        return {
-            media_type: handler
-            for media_type, handler in self.iter_handlers()
-            if handler in bound_components or handler in bound_component_types
-        }
+
+        bound_handlers = dict()
+        for media_type, handler in self.iter_handlers():
+            if handler in bound_components or handler in bound_component_types:
+                if bound_handlers.get(media_type):
+                    raise AlreadyRegisteredError(
+                        "Handler {} already registered for media type: {}".format(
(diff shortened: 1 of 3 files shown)


microcosm_pubsub/registry.py
```diff
@@ -109,11 +109,19 @@ def compute_bound_handlers(self, bindings):
...
+        for media_type, handler in self.iter_handlers():
+            if handler in bound_components or handler in bound_component_types:
+                if bound_handlers.get(media_type):
+                    raise AlreadyRegisteredError(
+                        "Handler {} already registered for media type: {}".format(
+                            bound_handlers.get(media_type).__name__,
+                            media_type,
+                        )
+                    )
+                bound_handlers[media_type] = handler
```
GabeLambda: There was a comment last time about making this a stream and still using the dict comprehension. I couldn't think of a nice way to do that without essentially re-doing the already-exists computation that I'm doing here.

microcosm_pubsub/tests/test_registry.py
```diff
@@ -104,12 +113,28 @@ def test_find_not_found(self):
...
         """
-        No exception raises when re-registering ab existing handler.
+        No exception raises when re-registering an existing handler.
 
         """
         self.registry.register(DerivedSchema.MEDIA_TYPE, noop_handler)
 
-    def test_regiser_another(self):
+    def test_duplicate_bound_handler_found(self):
+        self.graph.unlock()
```
GabeLambda: Explicitly forcing the graph to have a new handler without (I hope) affecting the other tests.

PR #73, opened 2018-05-04
Add reason parameter to Nack

sethisernhagen:
- add reason parameter to Nack

Why?

Eliminates need to log an extra event noting the nack reason.

### DIFF
--- a/microcosm_pubsub/errors.py
+++ b/microcosm_pubsub/errors.py
@@ -13,7 +13,9 @@ class Nack(Exception):
     An error that causes the current message to be nacked.
 
     """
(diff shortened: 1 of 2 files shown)


microcosm_pubsub/errors.py
```diff
@@ -13,7 +13,9 @@ class Nack(Exception):
     An error that causes the current message to be nacked.
 
     """
-    def __init__(self, visibility_timeout_seconds):
+    def __init__(self,  visibility_timeout_seconds, reason=None, extra=None):
+        super().__init__(reason)
```
jessemyers: What happens if `reason` is `None`?

PR #81, opened 2018-06-25
Fail gracefully if matched url is not found

saivickna:
Why?

Sometimes the associated URI with a pubsub message may not resolve to an endpoint in the service. We should simply return none for the route value in case this occurs.

### DIFF
--- a/microcosm_pubsub/producer.py
+++ b/microcosm_pubsub/producer.py
@@ -333,8 +333,8 @@ def configure_sns_producer(graph):
         opaque = None
 
     try:
-        register = bool(graph.publish_info_convention)
-    except NotBoundError:
+        register = bool(graph._cache.__getitem__('publish_info_convention'))
+    except KeyError:
         register = False
 
     if graph.config.sns_producer.skip is None:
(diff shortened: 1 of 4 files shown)


pelegdvir:
Can you explain why it is not a problem? I'm not sure that hiding the error is *always* the right thing.

takumig:
I just wanted to confirm that microcosm-pubsub was the culprit for cricle breakage.  I'll roll this back.
Once pubsub fix is released, this will build successfully: https://github.com/globality-corp/microcosm-pubsub/pull/81
@pelegdvir

takumig:
I agree that levant should not be publishing bogus uri.  Will fix this in levant.
But it's not introspection's place to bring down a service because of a condition it finds unkosher.  It's good to be rigorous, but introspection is not the place to do it.   If we want to enforce this sort of thing, we should write a validator.    So I still vote for this fix, aside from above proper fix. @pelegdvir

takumig:
If you insist on introspection taking down and endpoint, it should be emit a clear error message, rather than just throwing an exception.

microcosm_pubsub/producer.py
```diff
@@ -333,8 +333,8 @@ def configure_sns_producer(graph):
         opaque = None
 
     try:
-        register = bool(graph.publish_info_convention)
-    except NotBoundError:
+        register = bool(graph._cache.__getitem__('publish_info_convention'))
```
takumig: So this tests bound-but-not-created?  
It's pretty subtle.  A comment for future maintainer would be nice.

PR #85, opened 2018-07-06
@publish decorator

pelegdvir:
Simple method decorator that produces (sns) messages. Supports some customization.
Should be used in crud functions. Can also be used in handlers.

Basic examples:
```
    def __init__(self, graph):
        self.ns = Namespace(subject="company")
        self.identifier_key = company_id

    @publish()
    def create(**kwargs):
        return company(id="ID")
    produce:
        application/vnd.globality.pubsub._.created.company message
        with: {'uri': '/api/v1/company/ID'}

    @publish(media_type=changed("Company"))
    def create(**kwargs):
        return company(id="ID")
    produce:
        application/vnd.globality.pubsub._.changed.company message
        with: {'uri': '/api/v1/company/ID'}

    @publish(uri_string_args={"company_id": lambda company: company.id, "limit": lambda company: 1})
        return company(id="ID")
    produce:
        application/vnd.globality.pubsub._.created.company message
        with: {'uri': '/api/v1/company/ID?limit=1'}

```
Advanced examples are in the test file

What's next:
* Make the decorator introspectable
* Conditions (publish only if `model != None or model.event_type = "user_invited"`)  

Maybe:
* The customization lambda  functions can only access the "result" (for crud resolvers - the returned model) and not the graph / controller ("self") - it's on purpose. We want to make those resolvers simple. We can revisit this decision.
* >1 message support support (we can add two decorators to the same function, but a single decorator will publish a single message). I'm not sure if it's a good idea.


microcosm_pubsub/decorators/publish_decorators.py
```diff
@@ -0,0 +1,138 @@
...
+
+    @publish(media_type=changed("Company"))
+    def create(**kwargs):
+        return company(id="ID")
+    produce:
+        application/vnd.globality.pubsub._.changed.company message
+        with: {'uri': '/api/v1/company/ID'}
+
+    @publish(uri_string_args={"company_id": lambda company: company.id, "limit": lambda company: 1})
+        return company(id="ID")
```
saivickna: prolly missing a `def create(**kwargs)` here?

microcosm_pubsub/decorators/publish_decorators.py
```diff
@@ -0,0 +1,138 @@
...
+    :ns              - If ns is set to true or if media_type is unspecified,
+                       allows to specify any ns. Useful if component.ns is not set.
+    :identifier_key  - If default_uri is set to true, allows to specify any identifier_key.
+                       The value will still be model.id
+                       Useful if component.identifier_key is not set or when setting the ns arg.
+    :operation       - microcosm_flask.operations.Operation to use for autogenerating the default_uri.
+                       Default is Operation.Retrieve.
+                       It is recommended to set both operation and uri_string_args toghether.
+
+    """
```
wheremiah: Good doc string

microcosm_pubsub/decorators/publish_decorators.py
```diff
@@ -0,0 +1,138 @@
...
+        return company(id="ID")
+    produce:
+        application/vnd.globality.pubsub._.changed.company message
+        with: {'uri': '/api/v1/company/ID'}
+
+    @publish(uri_string_args={"company_id": lambda company: company.id, "limit": lambda company: 1})
+        return company(id="ID")
+    produce:
+        application/vnd.globality.pubsub._.created.company message
+        with: {'uri': '/api/v1/company/ID?limit=1'}
```
saivickna: I'd include an example with the Operation (maybe like the test case where you didn't include uri args).

microcosm_pubsub/decorators/publish_decorators.py
```diff
@@ -0,0 +1,138 @@
+"""
+Fluent decorators for methods.
+
+"""
+
+from marshmallow import ValidationError
+from functools import wraps
+from microcosm_flask.operations import Operation
```
jessemyers: This is not ok with me. We do not want to have pubsub depend on flask.

microcosm_pubsub/tests/test_publish_decorator.py
```diff
@@ -0,0 +1,198 @@
...
+    def test_set_message_params(self):
+        decorator = publish(media_type=DuckTypeSchema.MEDIA_TYPE, message_params={"quack": lambda model: model.name})
+        with self.graph.app.test_request_context():
+            create_decorated_controller(self.graph, decorator).retrieve()
+        assert_that(loads(self.graph.sns_producer.sns_client.publish.call_args[1]["Message"]), is_(equal_to({
+            "quack": "Name",
+        })))
+
+    def test_set_operation(self):
+        decorator = publish(operation=Operation.Search)
```
saivickna: I think it might be smart enough but how does this handle one of the RetrieveFor/SearchFor operations?

microcosm_pubsub/tests/test_publish_decorator.py
```diff
@@ -0,0 +1,198 @@
...
+        self.graph = create_object_graph("example", testing=True, loader=loader)
+        self.graph.use(
+            "configure_company_v1",
+            "configure_user_v1",
+            "sns_producer",
+        )
+
+        self.graph.lock()
+        self.graph.flask.test_client()
+        self.client = self.graph.flask.test_client()
```
wheremiah: Do you need the previous line?

microcosm_pubsub/decorators/publish_decorators.py
```diff
@@ -0,0 +1,138 @@
...
+
+    """
+    def decorator(func):
+        if default_uri and message_params and "uri" in message_params:
+            raise ValidationError("Cannot pass both uri extractor and while default_uri set to true")
+        operation_ = operation or Operation.Retrieve
+
+        @wraps(func)
+        def decorate(component, *func_args, **func_kwargs):
+            sns_producer = getattr(component, producer_key, None)
```
jessemyers: It's possible to:

```
getattr(component, producer_key, getattr(component.graph, producer_key, None)
```

microcosm_pubsub/decorators/publish_decorators.py
```diff
@@ -0,0 +1,138 @@
...
+            if sns_producer is None:
+                sns_producer = getattr(component.graph, producer_key)
+            media_type_to_publish = media_type or _create_default_media_type(component, **kwargs)
+
+            extractors = dict(message_params) if message_params else dict()
+            if default_uri:
+                uri = _create_default_uri_extractor(component, operation_, uri_string_args, **kwargs)
```
jessemyers: I think you want to have a `uri_factory` or `uri_extractor` be injected into this function and have `microcosm-flask` define an implementation that is decoupled from publishing.

microcosm_pubsub/tests/test_publish_decorator.py
pelegdvir: probably not

pelegdvir:
Splitting the PR to two prs: microcosm-flask & microcosm-pubsub

microcosm_pubsub/decorators/publish_decorators.py
pelegdvir: But then we assume that we have `self.graph` and it's not always the case. What we can do is:
```
getattr(component, producer_key, None) or getattr(component.graph, producer_key)
```

PR #86, opened 2018-07-06
@publish decorator

pelegdvir:
Lighter version of https://github.com/globality-corp/microcosm-pubsub/pull/85

Simple method decorator that produces (sns) messages. Supports some customization.

Simple example:
```
        @publish(
             media_type=created(company),
             uri=lambda component, model: f"/api/v1/company/{model.id}"
        )
        def create(**kwargs):
            return company(id="ID")
        produce:
            application/vnd.globality.pubsub._.created.company message
            with: {'uri': '/api/v1/company/ID'}
```


microcosm_pubsub/decorators/publish_decorators.py
```diff
@@ -0,0 +1,80 @@
...
+
+        @publish(media_type=created(company), company_id=lambda component, company: company.id)
+        def create(**kwargs):
+            return company(id="ID")
+        produce:
+            application/vnd.globality.pubsub._.created.company message
+            with: {'company_id': 'id'}
+
+    """
+    if media_type and media_type_extractor:
```
jessemyers: Seems like there's a possible simplification here where:
 - `media_type` is an `arg`, not a `kwarg` and is rerequired
 - we check if it is callable or not and call it if it is

microcosm_pubsub/decorators/publish_decorators.py
```diff
@@ -0,0 +1,80 @@
...
+        raise TypeError("Cannot pass both media_type and media_type_extractor")
+    if not media_type and not media_type_extractor:
+        raise TypeError("Must pass either media_type or media_type_extractor")
+
+    def decorator(func):
+        @wraps(func)
+        def decorate(component, *func_args, **func_kwargs):
+            sns_producer = getattr(component, producer_key, None) or getattr(component.graph, producer_key)
+            model = func(component, *func_args, **func_kwargs)
```
jessemyers: Maybe `result` instead of `model` as that makes some assumptions about the underlying function being a CRUD controller

microcosm_pubsub/decorators/publish_decorators.py
pelegdvir: I like it

pelegdvir:
Closing old PR - i'm not sure that we want to use this

PR #92, opened 2018-07-28
URIHandler.resource_type

pelegdvir:
New `URIHandler` property: `resource_type` - sets the returned resource type

### DIFF
--- a/microcosm_pubsub/handlers/uri_handler.py
+++ b/microcosm_pubsub/handlers/uri_handler.py
@@ -123,7 +123,7 @@ class URIHandler:
             raise Nack(self.nack_timeout)
         response.raise_for_status()
         body = response.json()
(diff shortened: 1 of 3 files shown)


microcosm_pubsub/handlers/uri_handler.py
```diff
@@ -118,7 +122,10 @@ def get_resource(self, message, uri):
         if response.status_code == codes.not_found and self.nack_if_not_found:
             raise Nack(self.nack_timeout)
         response.raise_for_status()
-        return response.json()
+        body = response.json()
+        if self.resource_type is dict:
```
jessemyers: Do L126-127 matter? If we ran `dict(**body)`, we'd have an equivalent value.

Or, better still, change the default resource type to be an identity function...

PR #100, opened 2018-08-05
Advanced context decorators

pelegdvir:
See "Please Fix Your Decorators": https://hynek.me/articles/decorators/

Handle 2 edge cases:
* Support methods decorations:
In python, we cannot `setattr` on instance methods (because they don't have `__dict__`)
For example:
```
class Example
    def func(self):
         pass

setattr(Example.func, 'x', 'x') # works
setattr(Example().func, 'x', 'x') # raises!
```
But we do want to use `binds` and `extracts` on them.
The solution was to wrap methods in functions. However, python `functools.wraps` hides the function signature - and causes `get_from_context` to fail. The solution for that is to use `inspect.signature` that has  instead of `inspect.getfullargspec`.

* Support multiple decorators for non functions / methods:
`functools.wraps` does not copy non functions / methods callable properties. That causes the first wrapper (`get_from_context`) to hide all other descriptors - such as `"_extracts"`.
```
@extract("x")
class Example
    def __call__(self, y):
         pass
```
We want to apply decorators on top of other decorators that are implemented as descriptors.
We achieve this using `functools.wraps` `assigned` optional argument. However - it requires that the `Chain` class will know the properties names.

New feature:
* `get_from_context` now supports default value function arguments (see `test_get_from_context_default_value`)

### DIFF
--- a/microcosm_pubsub/chain/chain.py
+++ b/microcosm_pubsub/chain/chain.py
@@ -40,7 +40,7 @@ class Chain:
         """
         Tuple naming attributes that self.context_decorators attributes should keep
         It should contain all attributes that set by relevant decorators
-        (such as @extracs)
+        (such as @extracts)
 
         """
         return DEFAULT_ASSIGNED
--- a/microcosm_pubsub/chain/chain.py
+++ b/microcosm_pubsub/chain/chain.py
@@ -4,6 +4,7 @@ from microcosm_pubsub.chain.context_decorators import (
     save_to_context,
     save_to_context_by_func_name,
     temporarily_replace_context_keys,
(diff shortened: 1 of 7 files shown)


microcosm_pubsub/chain/chain.py
```diff
@@ -34,6 +35,16 @@ def context_decorators(self):
             save_to_context_by_func_name,
         ]
 
+    @property
+    def context_decorators_assigned(self):
+        """
+        Tuple naming attributes that self.context_decorators attributes should keep
+        It should contain all attributes that set by relevant decorators
+        (such as @extracs)
```
jessemyers: `@extracts`

PR #105, opened 2018-08-21
Chain usability improvements.

jessemyers:
Make pubsub chains more usable (and more correct). Specifically:

 -  Ensure that `ForEachStatement` allows the same variable to be set in the context
    on each iteration by using an alternate version of the context that implements
    local scoping.

 -  Do not require `Chain` to be declared explicitly when it can be derived from
    statement context.

 -  Allow switch statement cases and try catch error cases to use a builder syntax.

 -  Allow variable extraction to be implemented using a more fluent builder style.

Example usage: https://github.com/globality-corp/frigg/pull/204/commits/924e387e78f24cc0ce09bcaf486d16826ad456a8

### DIFF
--- a/microcosm_pubsub/chain/chain.py
+++ b/microcosm_pubsub/chain/chain.py
@@ -63,18 +63,28 @@ class Chain:
         context = context or self.new_context_type()
         context.update(kwargs)
 
-        # Allow self reference
-        if "context" not in context:
-            context.update(context=context)
-        elif context["context"] is not context:
-            raise ValueError("context should be set as context")
-
         for link in self.links:
-            res = self.apply_decorators(context, link)()
+            func = self.apply_decorators(context, link)
+            res = func()
+
         return res
+
+    def __len__(self):
+        return len(self.links)
 
     def apply_decorators(self, context, link):
         decorated_link = link
         for decorator in self.context_decorators:
             decorated_link = decorator(context, decorated_link, self.context_decorators_assigned)
         return decorated_link
+
+    @classmethod
+    def make(cls, *args, chain=None, links=None, **kwargs):
+        if chain is not None:
+            return chain
+        elif links is not None:
+            return cls(*links)
+        elif args:
+            return cls(*args)
+        else:
+            raise ValueError("Must define one of 'chain', 'links', or '*args'")
--- a/microcosm_pubsub/chain/context.py
+++ b/microcosm_pubsub/chain/context.py
@@ -1,4 +1,8 @@
 from collections import MutableMapping
+from itertools import chain
+
+
+CONTEXT = "context"
 
 
 class SafeContext(MutableMapping):
@@ -6,21 +10,33 @@ class SafeContext(MutableMapping):
     Dictionary that raises on overwriting keys.
 
     """
-
     def __init__(self, *args, **kwargs):
         self.store = dict(*args, **kwargs)
 
     def __getitem__(self, key):
+        # enable context self-references
+        if key == CONTEXT:
+            return self
         return self.store[key]
 
     def __setitem__(self, key, value):
+        # do not allow overwrite of context self-references
+        if key == CONTEXT:
+            raise ValueError("May not reassign 'context' key")
+
         if key in self.store:
             raise ValueError(f"Key '{key}' already set")
+
         self.store[key] = value
 
     def assign(self, key, value):
+        # do not allow overwrite of context self-references
+        if key == CONTEXT:
+            raise ValueError("May not reassign 'context' key")
+
         # Skip validation
         self.store[key] = value
+        return self
 
     def __delitem__(self, key):
         del self.store[key]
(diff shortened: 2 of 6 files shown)


microcosm_pubsub/chain/chain.py
```diff
@@ -63,18 +63,28 @@ def __call__(self, context=None, **kwargs):
         context = context or self.new_context_type()
         context.update(kwargs)
 
-        # Allow self reference
```
jessemyers: Context self-reference is now implemented within the context types.

microcosm_pubsub/chain/chain.py
```diff
@@ -63,18 +63,28 @@ def __call__(self, context=None, **kwargs):
...
-        if "context" not in context:
-            context.update(context=context)
-        elif context["context"] is not context:
-            raise ValueError("context should be set as context")
-
         for link in self.links:
-            res = self.apply_decorators(context, link)()
+            func = self.apply_decorators(context, link)
```
jessemyers: Make the call chain more explicit here by separating the decoration and execution logic. I had a troublesome bug in my local context implementation that I mistakenly believed was part of the decorator logic because these steps were not separated.

microcosm_pubsub/chain/chain.py
```diff
@@ -63,18 +63,28 @@ def __call__(self, context=None, **kwargs):
...
+        return len(self.links)
+
     def apply_decorators(self, context, link):
         decorated_link = link
         for decorator in self.context_decorators:
             decorated_link = decorator(context, decorated_link, self.context_decorators_assigned)
         return decorated_link
+
+    @classmethod
+    def make(cls, *args, chain=None, links=None, **kwargs):
```
jessemyers: Shortcut for creating a chain from various arguments:
 - `chain` is the backwards compatible way
 - `links` gives a better name for a list of statements defined by a kwarg
 - `*args* is the most likely usage in the simplified syntax

microcosm_pubsub/chain/context.py
```diff
@@ -33,3 +49,41 @@ def __len__(self):
...
+
+    def local(self, *args, **kwargs):
+        """
+        Create a locally scoped child context.
+
+        """
+        return ScopedSafeContext(self, *args, **kwargs)
+
+
+class ScopedSafeContext(SafeContext):
```
jessemyers: This class implements a scoped context. It defers to the parent when it does not have a value itself.

We do NOT currently allow variable shadowing in the local scope.

microcosm_pubsub/chain/context.py
```diff
@@ -33,3 +49,41 @@ def __len__(self):
 
     def __getattr__(self, key):
         return self[key]
+
+    def local(self, *args, **kwargs):
```
jessemyers: Create a local scope from this one.

microcosm_pubsub/chain/statements.py
```diff
@@ -1,37 +1,59 @@
+from microcosm_pubsub.chain import Chain
+
+
 class ArgumentExtractor:
-    def __init__(self, name, key, key_property=None):
-        self.name = name
-        self.key = key
-        self.key_property = key_property
+    """
```
jessemyers: Reimplemented to:
 - allow `this` to use dotted syntax: `foo.bar`
 - allow `that` to be defined "later" via the `as_` builder

That is:
```
extract("foo", "bar", "baz")
``
can now be defined as:
```
assign("bar.baz").as_("foo")
```

File: microcosm_pubsub/chain/statements.py [hunk 7]
```diff
@@ -1,37 +1,59 @@
...
+
+    """
     def __init__(self, key):
         self.key = key
         self._then = None
         self._otherwise = None
 
-    def then(self, chain):
-        self._then = chain
+    def then(self, *args, **kwargs):
```
jessemyers: Most of the chain inputs now call `Chain.make` to allow simplified inputs.

microcosm_pubsub/chain/statements.py
```diff
@@ -54,12 +90,15 @@ def __init__(self, key):
     def __str__(self):
         return f"switch_{self.key}"
 
-    def case(self, key, chain):
-        self._cases[key] = chain
+    def case(self, key, *args, **kwargs):
```
jessemyers: If we have a `key` but not a chain, defer using a builder.

microcosm_pubsub/chain/statements.py
```diff
@@ -70,26 +109,29 @@ def __call__(self, context):
...
-    def __init__(self, chain):
-        self.chain = chain
+    def __init__(self, *args, **kwargs):
+        self.chain = Chain.make(*args, **kwargs)
         self._cases = dict()
         self._otherwise = None
 
-    def catch(self, key, chain):
-        self._cases[key] = chain
+    def catch(self, key, *args,  **kwargs):
```
jessemyers: If we have a `key` but not a chain, defer using a builder.

PR #110, opened 2018-08-22
All filtering in for_each

jessemyers:


microcosm_pubsub/tests/chain/statements/test_for_each.py
```diff
@@ -36,3 +36,16 @@ def test_for_each_simplified():
         chain(items=["a", "b", "c"]),
         is_(equal_to(["C", "B", "A"])),
     )
+
```
GabeLambda: Maybe a test for testing a filter func in `when()`?

GabeLambda: Then again, it's tested implicitly in `when_not_none`.

PR #112, opened 2018-08-27
Enable assigning constants and function values

jessemyers:


microcosm_pubsub/tests/chain/statements/test_assign.py
```diff
@@ -91,3 +93,23 @@ def test_assign_attribute():
...
+    chain = Chain(
+        assign_constant("arg").to("param"),
+    )
+    assert_that(
+        chain(),
+        is_(equal_to("arg")),
+    )
+
+
+def test_assign_function():
```
jessemyers: I added this so we can do something like:

```
Chain(
    assign_function(lambda context: dict()).to("my dict"),
)
```

PR #113, opened 2018-09-11
X-Request-TTL support

pelegdvir:
All handlers now support `X-Request-TTL` field:
* If not set - sets it to a configurable number
* Always decrease the number
* If gets a message with `X-Request-TTL: 0` - raises

Why?
* Better logs
* Kill infinite handler loops

Details:
* Updates the TTL when receiving a message - not on sending
* Configuration - i'm not so sure about that - i've used `graph.config.sqs_message_context.initial_ttl` - even that we used it in the dispatcher

### DIFF
--- a/microcosm_pubsub/context.py
+++ b/microcosm_pubsub/context.py
@@ -3,7 +3,7 @@ Message context.
 
 """
 
-from microcosm.api import defaults
+from microcosm.api import defaults, typed
 
 
 def sqs_message_context(message_dct, **kwargs):
@@ -18,8 +18,8 @@ def sqs_message_context(message_dct, **kwargs):
 
 
 @defaults(
-    enable_ttl=True,
-    initial_ttl=32,
+    enable_ttl=typed(bool, default_value=True),
+    initial_ttl=typed(int, default_value=32),
 )
 def configure_sqs_message_context(graph):
     """
--- a/microcosm_pubsub/context.py
+++ b/microcosm_pubsub/context.py
@@ -3,6 +3,8 @@ Message context.
 
 """
 
+from microcosm.api import defaults
+
 
 def sqs_message_context(message_dct, **kwargs):
     context = message_dct.get("opaque_data", dict())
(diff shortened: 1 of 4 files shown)


microcosm_pubsub/context.py
```diff
@@ -15,6 +17,10 @@ def sqs_message_context(message_dct, **kwargs):
     return context
 
 
+@defaults(
+    enable_ttl=True,
```
jessemyers: You should use `<key>=typed(<type>, default_value=<value>)` here; otherwise environment variables will not be converted properly from strings.

microcosm_pubsub/dispatcher.py
```diff
@@ -26,6 +26,8 @@ def __init__(self, graph):
         self.sqs_consumer = graph.sqs_consumer
         self.sqs_message_context = self._find_sqs_message_context(graph)
         self.sqs_message_handler_registry = graph.sqs_message_handler_registry
+        self.enable_ttl = graph.config.sqs_message_context.enable_ttl
```
pelegdvir: note: dispatcher is using `graph.config.sqs_message_context` config

microcosm_pubsub/context.py
```diff
@@ -15,6 +17,10 @@ def sqs_message_context(message_dct, **kwargs):
     return context
 
 
+@defaults(
+    enable_ttl=typed(bool, default_value=True),
```
jessemyers: You probably want to use our custom [boolean](https://github.com/globality-corp/microcosm/blob/develop/microcosm/config/types.py#L8) to properly handle strings

microcosm_pubsub/dispatcher.py
```diff
@@ -83,6 +85,17 @@ def handle_message(self, message, bound_handlers):
                 handler=titleize(handler.__class__.__name__),
                 uri=content.get("uri"),
             ))
+
+            if self.enable_ttl:
+                ttl = content.get("X-Request-TTL", self.initial_ttl) - 1
+                if ttl == -1:
+                    self.logger.warning(
+                        f"Error handling SQS message: {media_type} - TTL expired",
+                        extra=extra
```
jessemyers: trailing comma

PR #115, opened 2018-09-25
X-Request-Ttl to opaque

pelegdvir:
Few changes:
* Rename `X-Request-TTL` to `X-Request-Ttl` - to be compatible to other microcosm packages
*  Store `X-Request-Ttl` in graph.opaque instead of message context - standard convention
*  Pass `X-Request-Ttl` as a string (because the retrieved message contains strings)

### DIFF
--- a/microcosm_pubsub/context.py
+++ b/microcosm_pubsub/context.py
@@ -5,17 +5,45 @@ Message context.
 
 from microcosm.api import defaults, typed
 from microcosm.config.types import boolean
+from microcosm_logging.decorators import logger
+
+from microcosm_pubsub.errors import TTLExpired
 
 
-def sqs_message_context(message_dct, **kwargs):
-    context = message_dct.get("opaque_data", dict())
+@logger
+class SQSMessageContext:
+    """
+    The message context controls what data you want to associate
+    with your daemon handler context, e.g. some combination of keys from the message or
+    additional metadata fetched from elsewhere.
 
-    # If there is a uri add it to the context
-    if message_dct.get("uri"):
-        context["uri"] = message_dct.get("uri")
+    """
+    def __init__(self, graph):
+        self.enable_ttl = graph.config.sqs_message_context.enable_ttl
+        self.initial_ttl = graph.config.sqs_message_context.initial_ttl
 
-    context.update(**kwargs)
-    return context
+    def __call__(self, message_dct, **kwargs):
+        context = message_dct.get("opaque_data", dict())
+
+        # If there is a uri add it to the context
+        if message_dct.get("uri"):
+            context["uri"] = message_dct.get("uri")
+
+        if self.enable_ttl:
+            try:
+                ttl = int(context["X-Request-Ttl"])
+            except KeyError:
+                ttl = self.initial_ttl
+            if ttl == 0:
+                self.logger.warning(
+                    "Error handling SQS message - TTL expired",
+                    extra=context,
+                )
+                raise TTLExpired(extra=context)
+            context["X-Request-Ttl"] = str(ttl - 1)
+
+        context.update(**kwargs)
+        return context
 
 
 @defaults(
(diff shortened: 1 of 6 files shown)


microcosm_pubsub/dispatcher.py
```diff
@@ -133,7 +113,7 @@ def invoke_handler(self, handler, media_type, content):
                 ),
                 extra=self.sqs_message_context(content)
             )
-            raise TTLExpired()
```
pelegdvir: Revert an accidental change

microcosm_pubsub/context.py
```diff
@@ -5,17 +5,45 @@
...
+            context["uri"] = message_dct.get("uri")
+
+        if self.enable_ttl:
+            try:
+                ttl = int(context["X-Request-Ttl"])
+            except KeyError:
+                ttl = self.initial_ttl
+            if ttl == 0:
+                self.logger.warning(
+                    f"Error handling SQS message - TTL expired",
```
GabeLambda: Nit: don't need format-string here.

microcosm_pubsub/context.py
```diff
@@ -5,17 +5,45 @@
...
+                    extra=context,
+                )
+                raise TTLExpired(extra=context)
+            context["X-Request-Ttl"] = str(ttl - 1)
```
GabeLambda: What's this for?

microcosm_pubsub/context.py
```diff
@@ -24,12 +52,10 @@ def sqs_message_context(message_dct, **kwargs):
 )
 def configure_sqs_message_context(graph):
     """
-    Configure the message context function which controls what data you want to associate
-    with your daemon handler context, e.g. some combination of keys from the message or
-    additional metadata fetched from elsewhere.
+    Configure the message context object.
```
GabeLambda: Why remove the comment here?

microcosm_pubsub/dispatcher.py
```diff
@@ -24,19 +23,11 @@ class SQSMessageDispatcher:
     def __init__(self, graph):
         self.opaque = graph.opaque
         self.sqs_consumer = graph.sqs_consumer
-        self.sqs_message_context = self._find_sqs_message_context(graph)
+        self.sqs_message_context = graph.sqs_message_context
         self.sqs_message_handler_registry = graph.sqs_message_handler_registry
         self.enable_ttl = graph.config.sqs_message_context.enable_ttl
         self.initial_ttl = graph.config.sqs_message_context.initial_ttl
 
-    def _find_sqs_message_context(self, graph):
```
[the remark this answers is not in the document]
GabeLambda: Still not sure what case this was trying to control for here.

microcosm_pubsub/context.py
pelegdvir: Moved it to `SQSMessageContext`

microcosm_pubsub/dispatcher.py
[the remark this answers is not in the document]
pelegdvir: From git history - seems to be an old backward compatibility check

microcosm_pubsub/context.py
pelegdvir: the `str`?  because the retrieved message contains only strings (and not numbers)

GabeLambda: I mean the `- 1`.

microcosm_pubsub/context.py
```diff
@@ -5,17 +5,45 @@
...
 
-    context.update(**kwargs)
-    return context
+    """
+    def __init__(self, graph):
+        self.enable_ttl = graph.config.sqs_message_context.enable_ttl
+        self.initial_ttl = graph.config.sqs_message_context.initial_ttl
+
+    def __call__(self, message_dct, **kwargs):
+        context = message_dct.get("opaque_data", dict())
```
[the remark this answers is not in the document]
pelegdvir: Yes.
That's the function that we're passing to `opaque.initialize` - to initialize opaque with pubsub logic 
`with graph.opaque.initialize(graph.sqs_message_context, message, ...):`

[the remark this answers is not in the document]
jessemyers: OK, I missed that.

PR #121, opened 2018-10-15
Initial refactor for pubsub dispatcher

jessemyers:
The inner layers of the pubsub dispatcher combine flow control, context management, logging,
timing, and n(ack) results confusingly. This is a first pass at cleaning up this logic,
with an eye towards better measurement of behavior.

Also: embrace `unittest.mock`

### DIFF
--- a/microcosm_pubsub/daemon.py
+++ b/microcosm_pubsub/daemon.py
@@ -86,8 +86,8 @@ class ConsumerDaemon(Daemon):
         Implement daemon by sinking messages from the consumer to a dispatcher function.
 
         """
-        result = graph.sqs_message_dispatcher.handle_batch(self.bound_handlers)
-        if not result.message_count:
+        results = graph.sqs_message_dispatcher.handle_batch(self.bound_handlers)
+        if not results:
             raise SleepNow()
 
     @classmethod
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -2,16 +2,14 @@
 Process batches of messages.
 
 """
-from collections import namedtuple
 from inflection import titleize
+from typing import List
 
 from microcosm_logging.decorators import context_logger, logger
-
-from microcosm_pubsub.errors import Nack, SkipMessage, TTLExpired
 from microcosm_logging.timing import elapsed_time
 
-
-DispatchResult = namedtuple("DispatchResult", ["message_count", "error_count", "ignore_count"])
+from microcosm_pubsub.errors import Nack, SkipMessage, TTLExpired
+from microcosm_pubsub.result import MessageHandlingResult
 
 
 @logger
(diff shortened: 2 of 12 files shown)


microcosm_pubsub/daemon.py
```diff
@@ -86,8 +86,8 @@ def __call__(self, graph):
         Implement daemon by sinking messages from the consumer to a dispatcher function.
 
         """
-        result = graph.sqs_message_dispatcher.handle_batch(self.bound_handlers)
-        if not result.message_count:
+        results = graph.sqs_message_dispatcher.handle_batch(self.bound_handlers)
```
jessemyers: Instead of return a structure with a `message_count` (which is unused), we return a list of results (with data we should be able to use for metrics)

microcosm_pubsub/result.py
```diff
@@ -0,0 +1,57 @@
...
+                result = MessageHandlingResultType.SKIPPED
+            message.ack()
+        except SkipMessage:
+            result = MessageHandlingResultType.SKIPPED
+            message.ack()
+        except Nack as nack:
+            result = MessageHandlingResultType.RETRIED
+            message.nack(nack.visibility_timeout_seconds)
+        except Exception as error:
+            result = MessageHandlingResultType.FAILED
```
GabeLambda: Technically speaking, this will also be retried (FAILED might indicate that this was a terminal state for this event).

jessemyers: Better naming suggestions?

microcosm_pubsub/dispatcher.py
```diff
@@ -34,21 +31,14 @@ def handle_batch(self, bound_handlers):
 
         :returns: a `DispatchResult` with metrics around message consumption
```
GabeLambda: Nit: doc now out of date.

microcosm_pubsub/result.py
jessemyers: How about: `DELAYED` vs `FAILED`?

microcosm_pubsub/dispatcher.py
GabeLambda: Also, if we're embracing 3.7+ (I see direct dataclass import), might I suggest adding typehinting?

microcosm_pubsub/result.py
```diff
@@ -0,0 +1,57 @@
...
+
+from microcosm_pubsub.errors import Nack, SkipMessage
+
+
+@unique
+class MessageHandlingResultType(Enum):
+    # Messaging handling failed.
+    FAILED = "FAILED"
+
+    # Messaging handling was intentially retried.
```
GabeLambda: intentionally

microcosm_pubsub/result.py
GabeLambda: DELAYED sounds closer to the RETRIED case, actually.

If the message context is available here, we *could* still use FAILED if we knew the current retry count, and if we'd stop trying after this point.

Still, FAILED is probably more readable than something like UNHANDLED_ERROR, and it has a similar connotation anyways.

jessemyers: Conclusion: no change here

microcosm_pubsub/tests/test_result.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+
+        assert_that(
+            result,
+            has_properties(
+                media_type="application/vnd.microcosm.derived",
+                result=MessageHandlingResultType.FAILED,
+            ),
+        )
+        # no nack (XXX)
```
jessemyers: The default backoff policy doesn't return a default timeout so we don't nack by default on error. This is not great.

PR #122, opened 2018-10-15
Use separate variable to control nack timeouts in URIHandlers

GabeLambda:
There are cases where handlers downstream will explicitly nack events
using `nack_timeout`, and will override the default value. This should
be a separate concern from the default nack we configure to control how
we handle 404s on resource retrieve.

### DIFF
--- a/microcosm_pubsub/handlers/uri_handler.py
+++ b/microcosm_pubsub/handlers/uri_handler.py
@@ -32,10 +32,15 @@ class URIHandler:
     """
     __metaclass__ = ABCMeta
 
-    def __init__(self, graph, nack_timeout=1, resource_nack_timeout=1):
+    def __init__(self, graph, retry_nack_timeout=1, resource_nack_timeout=1):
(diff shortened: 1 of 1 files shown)


microcosm_pubsub/handlers/uri_handler.py
```diff
@@ -32,9 +32,10 @@ class URIHandler:
     """
     __metaclass__ = ABCMeta
 
-    def __init__(self, graph, nack_timeout=1):
+    def __init__(self, graph, nack_timeout=1, resource_nack_timeout=1):
```
jessemyers: Suggestion: rename both timeout variables for clarity and use an `@property` for the legacy `nack_timeout` usage compatibility (until it is fixed)

GabeLambda: Sure.

GabeLambda:
This will break some downstream handlers (that set `nack_timeout` directly), but they need to be reworked anyways.

PR #125, opened 2018-10-16
Add EXPIRED and IGNORED result states.

jessemyers:
Move all message handling control flow through the same mechanism.


microcosm_pubsub/dispatcher.py
```diff
@@ -74,6 +65,18 @@ def handle_message(self, message, bound_handlers):
             )
             return result
 
+    def find_handler(self, message, bound_handlers):
+        media_type = message.media_type
+        content = message.content
+
+        if content is None:
```
jessemyers: The two error cases in this handler imply that we should not "count" the message at all (e.g. for metrics).

The second  check, at least, is a very common case because most daemons see a lot of message they do not process. It should not be treated the same as `SkipMessage`

microcosm_pubsub/result.py
```diff
@@ -41,9 +47,15 @@ def invoke(cls, func, message, **kwargs):
             else:
                 result = MessageHandlingResultType.SKIPPED
             message.ack()
+        except IgnoreMessage:
+            result = MessageHandlingResultType.IGNORED
+            message.ack()
         except SkipMessage:
             result = MessageHandlingResultType.SKIPPED
             message.ack()
+        except TTLExpired:
```
jessemyers: This case would have been handled as `FAILED` previously (e.g. `nack` not `ack` and retry)

microcosm_pubsub/errors.py
```diff
@@ -22,7 +29,21 @@ def __repr__(self):
...
     """
-    Control-flow exception to skip resource processing.
+    Stop processing a message and do not retry processing.
+
+    Indicates that message content failed a precondition within a handler.
+
+    """
+    def __init__(self, reason, extra=None):
+        super().__init__(reason)
+        self.extra = extra or dict()
```
pelegdvir: nice

microcosm_pubsub/dispatcher.py
pelegdvir: I don't see why the first error case should be in `find_handler` function. It should live in a separate assertion function

PR #126, opened 2018-10-16
Simplify usage of opaque context in message dispatcher

jessemyers:
The opaque context is meant to capture generic state during program execution and
propagate to downstream services (via swagger or SNS calls). The dispatcher uses
this context in numerous ways and has let these usages get out-of-control:

 1. The dispatcher initializes a new context once it has enough information to
    not `IGNORE` a message

 2. The dispatcher records timing data within this context.

 3. The dispatcher explicitly passes `extra` to its logger using this context.

 4. The dispatcher overwrites handler loggers to use this context as their `extra`

This PR brings together the first three cases (but no yet the fourth) and lays
groundwork for recording timing data.

### DIFF
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -2,9 +2,10 @@
 Process batches of messages.
 
 """
-from inflection import titleize
-from typing import List
+from contextlib import contextmanager
+from typing import Dict, List, Tuple
 
+from inflection import titleize
 from microcosm_logging.decorators import context_logger, logger
 from microcosm_logging.timing import elapsed_time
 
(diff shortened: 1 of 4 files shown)


microcosm_pubsub/dispatcher.py
```diff
@@ -30,40 +31,57 @@ def handle_batch(self, bound_handlers) -> List[MessageHandlingResult]:
 
         """
         return [
-            MessageHandlingResult.invoke(
-                func=self.handle_message,
-                message=message,
-                bound_handlers=bound_handlers
-            )
+            self.handle_or_ignore_message(message, bound_handlers)
```
jessemyers: I am slowly moving the `MessageHandlingResult` inward so that we can capture more information (e.g. timing) for message handling. This necessarily involves moving the message context (e.g. `opaque`) outward.

PR #127, opened 2018-10-16
Simplify usage of opaque context in message dispatcher

jessemyers:
The opaque context is meant to capture generic state during program execution and
propagate to downstream services (via swagger or SNS calls). The dispatcher uses
this context in numerous ways and has let these usages get out-of-control:

 1. The dispatcher initializes a new context once it has enough information to
    not `IGNORE` a message

 2. The dispatcher records timing data within this context.

 3. The dispatcher explicitly passes `extra` to its logger using this context.

 4. The dispatcher overwrites handler loggers to use this context as their `extra`

This PR brings together the first three cases (but no yet the fourth) and lays
groundwork for recording timing data.

### DIFF
--- a/microcosm_pubsub/context.py
+++ b/microcosm_pubsub/context.py
@@ -2,15 +2,14 @@
 Message context.
 
 """
+from typing import Dict
 
 from microcosm.api import defaults, typed
 from microcosm.config.types import boolean
 from microcosm_logging.decorators import logger
 
-from microcosm_pubsub.errors import TTLExpired
-
-
-TTL_KEY = "X-Request-Ttl"
+from microcosm_pubsub.constants import TTL_KEY, URI_KEY
+from microcosm_pubsub.message import SQSMessage
 
 
 @defaults(
@@ -27,43 +26,39 @@ class SQSMessageContext:
         self.enable_ttl = graph.config.sqs_message_context.enable_ttl
         self.initial_ttl = graph.config.sqs_message_context.initial_ttl
 
-    def __call__(self, message, **kwargs):
+    def __call__(self, context, **kwargs) -> Dict[str, str]:
         """
         Create a new context from a message.
 
         """
-        # start with opaque data passed in message
-        if "opaque_data" in message:
-            context = message["opaque_data"].copy()
-        else:
-            context = dict()
+        if isinstance(context, SQSMessage):
+            return self.from_sqs_message(context, **kwargs)
 
-        # merge in explicit and derived arguments
-        context.update(
-            **self.uri_for(context, message),
-            **self.ttl_for(context, message),
+        # XXX we want to remove this
+        return self.from_dict(context, **kwargs)
+
+    def from_sqs_message(self, message, **kwargs):
+        context = dict(
+            # include the message id
+            message_id=message.message_id,
+            # include
+            **message.opaque_data,
             **kwargs,
         )
 
+        # include the TTL (if enabled)
+        if self.enable_ttl:
+            ttl = message.ttl if message.ttl is not None else self.initial_ttl
+            context[TTL_KEY] = str(ttl - 1)
+
+        # include the URI (if there is one)
+        if message.uri:
+            context[URI_KEY] = message.uri
+
         return context
 
-    def uri_for(self, context, message):
-        uri = message.get("uri")
-        return dict(uri=uri) if uri else dict()
-
-    def ttl_for(self, context, message):
-        try:
-            ttl = int(context[TTL_KEY])
-        except KeyError:
-            ttl = self.initial_ttl
-
-        if ttl == 0:
-            self.logger.warning(
-                "Error handling SQS message - TTL expired",
-                extra=context,
-            )
-            raise TTLExpired(extra=context)
-
-        return {
-            TTL_KEY: str(ttl - 1),
-        }
+    def from_dict(self, dct, **kwargs):
+        return dict(
+            **dct.get("opaque_data", {}),
+            **kwargs,
+        )
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -2,13 +2,14 @@
 Process batches of messages.
 
 """
-from inflection import titleize
 from typing import List
 
+from inflection import titleize
 from microcosm_logging.decorators import context_logger, logger
 from microcosm_logging.timing import elapsed_time
 
-from microcosm_pubsub.errors import IgnoreMessage, Nack, SkipMessage
+from microcosm_pubsub.context import TTL_KEY
+from microcosm_pubsub.errors import IgnoreMessage, Nack, SkipMessage, TTLExpired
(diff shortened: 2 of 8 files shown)


microcosm_pubsub/context.py
```diff
@@ -27,43 +26,39 @@ def __init__(self, graph):
         self.enable_ttl = graph.config.sqs_message_context.enable_ttl
         self.initial_ttl = graph.config.sqs_message_context.initial_ttl
 
-    def __call__(self, message, **kwargs):
+    def __call__(self, context, **kwargs) -> Dict[str, str]:
```
jessemyers: The `SQSMessageContext` now supports passing in an `SQSMessage` directly. 

This works as long as we don't also invoke using a `dict`, which we do in two places. I'll point those out and fix in a future PR.

microcosm_pubsub/context.py
```diff
@@ -27,43 +26,39 @@ def __init__(self, graph):
...
             **kwargs,
         )
 
-        return context
-
-    def uri_for(self, context, message):
-        uri = message.get("uri")
-        return dict(uri=uri) if uri else dict()
+        # include the TTL (if enabled)
+        if self.enable_ttl:
```
jessemyers: I am no longer validating TTL on context load because:
 - Our `elapsed_time` context manager saves its output in `self.opaque`
 - We want our `MessageHandlingResult` to record timing data (for metrics reporting)

So: `self.opaque` needs to be initialized outside of the result invocation (or we can't capture the reporting data).

But:

 - We want our `MessageHandlingResult` to capture all message handling errors (for result reporting)

So: we can't have errors raised in the `self.opaque` initialization

microcosm_pubsub/dispatcher.py
```diff
@@ -30,40 +31,60 @@ def handle_batch(self, bound_handlers) -> List[MessageHandlingResult]:
...
+        # initialize a context for the current message
+        with self.opaque.initialize(self.sqs_message_context, message):
+            # save elapsed time information to the context
+            with elapsed_time(self.opaque):
+                message_handling_result = MessageHandlingResult.invoke(
+                    func=self._handle_message,
+                    message=message,
+                    bound_handlers=bound_handlers,
+                )
+            message_handling_result.elapsed_time = self.opaque["elapsed_time"]
```
jessemyers: Notice how this nesting works. We set `elapsed_time` outside of the inner nesting but inside the outer one.

microcosm_pubsub/dispatcher.py
```diff
@@ -30,40 +31,59 @@ def handle_batch(self, bound_handlers) -> List[MessageHandlingResult]:
...
+
+    def _handle_message(self, message, bound_handlers) -> bool:
+        """
+        Handle a single message.
 
+        :raises: Nack, IgnoreMessage, SkipMessage, TTLExpired
+
+        """
+        # we might just want to ignore the message
+        self.validate_ttl()
```
jessemyers: Validate TTL here.

microcosm_pubsub/dispatcher.py
```diff
@@ -86,7 +106,9 @@ def find_handler(self, message, bound_handlers):
 
         """
         try:
-            return self.sqs_message_handler_registry.find(message.media_type, bound_handlers)
+            handler = self.sqs_message_handler_registry.find(message.media_type, bound_handlers)
+            self.opaque["handler"] = titleize(handler.__class__.__name__)
```
jessemyers: Save handling information once we have it. We cannot do this during initialization of `self.opaque` since it would allow an error to be raised before the result handling (if there was no such handler)

microcosm_pubsub/dispatcher.py
```diff
@@ -105,19 +127,24 @@ def invoke_handler(self, handler, media_type, content):
             )
             return handler_with_context(content)
         except SkipMessage as skipped:
-            extra = self.sqs_message_context(content)
-            extra.update(skipped.extra)
+            self.opaque.update(skipped.extra)
             logger.info(
                 "Skipping message for reason: {}".format(str(skipped)),
-                extra=extra,
+                extra=self.opaque.as_dict(),
```
jessemyers: I am going to work on the logging code next. I'm not clear whether `extra=` requires a mapping or a dict.

microcosm_pubsub/dispatcher.py
```diff
@@ -105,19 +127,24 @@ def invoke_handler(self, handler, media_type, content):
             )
             return handler_with_context(content)
```
jessemyers: The `handler_with_context` is passed `self.message_context`... and invokes it with a dictionary.

I barely understand what this code is trying to do so I haven't fixed it yet.

microcosm_pubsub/handlers/uri_handler.py
```diff
@@ -125,7 +125,8 @@ def get_resource(self, message, uri):
         Passes message context.
 
         """
-        headers = self.sqs_message_context(message)
+        # XXX we only want X-Request headers
```
jessemyers: I don't think we need to use `sqs_mexsage_context` here, but we also don't have good tests for this.

microcosm_pubsub/context.py
GabeLambda: Nit: annotate function args (this can explicitly label it as a `Union` for now, if you're not immediately fixing the `dict` case).

microcosm_pubsub/dispatcher.py
```diff
@@ -30,40 +31,59 @@ def handle_batch(self, bound_handlers) -> List[MessageHandlingResult]:
...
-                func=self.handle_message,
-                message=message,
-                bound_handlers=bound_handlers
-            )
+            self.handle_message(message, bound_handlers)
             for message in self.sqs_consumer.consume()
         ]
 
-    def handle_message(self, message, bound_handlers):
+    def handle_message(self, message, bound_handlers) -> MessageHandlingResult:
```
GabeLambda: Nit: param annotation

microcosm_pubsub/message.py
```diff
@@ -38,3 +39,21 @@ def nack(self, visibility_timeout_seconds=None):
...
+        return self.content.get("opaque_data", {})
+
+    @property
+    def ttl(self):
+        if TTL_KEY not in self.opaque_data:
+            return None
+        return int(self.opaque_data[TTL_KEY])
+
+    @property
+    def uri(self):
```
pelegdvir: !!!

microcosm_pubsub/dispatcher.py
```diff
@@ -30,40 +31,59 @@ def handle_batch(self, bound_handlers) -> List[MessageHandlingResult]:
...
-                "Handled {handler}",
-                extra=extra,
-            )
-            return result
+    def validate_ttl(self):
+        """
+        Validate that the current message is not expired.
+
+        :raises TTLExpired
+        """
```
pelegdvir: Don't we need to add `if not enable_ttl:    return`?

PR #128, opened 2018-10-16
Use opaque for URI handler headers.

jessemyers:
This is consistent with other (internal) usage for swagger clients.

### DIFF
--- a/microcosm_pubsub/handlers/uri_handler.py
+++ b/microcosm_pubsub/handlers/uri_handler.py
@@ -33,7 +33,7 @@ class URIHandler:
     __metaclass__ = ABCMeta
 
     def __init__(self, graph, retry_nack_timeout=1, resource_nack_timeout=1):
-        self.sqs_message_context = graph.sqs_message_context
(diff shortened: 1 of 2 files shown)


microcosm_pubsub/handlers/uri_handler.py
```diff
@@ -125,14 +125,20 @@ def get_resource(self, message, uri):
...
-        # XXX we only want X-Request headers
-        headers = self.opaque.as_dict()
+        headers = self.get_headers(message)
         response = get(uri, headers=headers)
         if response.status_code == codes.not_found and self.nack_if_not_found:
             raise Nack(self.resource_nack_timeout)
         response.raise_for_status()
         return response.json()
 
+    def get_headers(self, message):
```
GabeLambda: Is the signature here just to look like the call in our clients? `message` isn't used here.

jessemyers: I am anticipating a subclass wanting to extend this behavior using `message` content.

PR #129, opened 2018-10-16
Refactor dispatcher logging

jessemyers:
Make logging a function of the `MessageHandlingResult`

### DIFF
--- a/microcosm_pubsub/context.py
+++ b/microcosm_pubsub/context.py
@@ -26,16 +26,12 @@ class SQSMessageContext:
         self.enable_ttl = graph.config.sqs_message_context.enable_ttl
         self.initial_ttl = graph.config.sqs_message_context.initial_ttl
 
-    def __call__(self, context, **kwargs) -> Dict[str, str]:
+    def __call__(self, context: SQSMessage, **kwargs) -> Dict[str, str]:
         """
         Create a new context from a message.
 
         """
-        if isinstance(context, SQSMessage):
-            return self.from_sqs_message(context, **kwargs)
-
-        # XXX we want to remove this
-        return self.from_dict(context, **kwargs)
+        return self.from_sqs_message(context, **kwargs)
 
     def from_sqs_message(self, message, **kwargs):
         context = dict(
@@ -56,9 +52,3 @@ class SQSMessageContext:
             context[URI_KEY] = message.uri
 
         return context
-
-    def from_dict(self, dct, **kwargs):
-        return dict(
-            **dct.get("opaque_data", {}),
-            **kwargs,
-        )
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -9,7 +9,7 @@ from microcosm_logging.decorators import context_logger, logger
 from microcosm_logging.timing import elapsed_time
 
 from microcosm_pubsub.context import TTL_KEY
-from microcosm_pubsub.errors import IgnoreMessage, Nack, SkipMessage, TTLExpired
+from microcosm_pubsub.errors import IgnoreMessage, TTLExpired
 from microcosm_pubsub.result import MessageHandlingResult
 
 
@@ -49,7 +49,11 @@ class SQSMessageDispatcher:
                     message=message,
                     bound_handlers=bound_handlers,
                 )
+            # save elapsed time
             message_handling_result.elapsed_time = self.opaque["elapsed_time"]
+            # log handler outcome
+            logger = self.choose_logger(message.handler)
+            message_handling_result.log(logger, self.opaque)
         return message_handling_result
 
     def _handle_message(self, message, bound_handlers) -> bool:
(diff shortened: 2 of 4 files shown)


microcosm_pubsub/context.py
```diff
@@ -26,16 +26,12 @@ def __init__(self, graph):
...
-    def __call__(self, context, **kwargs) -> Dict[str, str]:
+    def __call__(self, context: SQSMessage, **kwargs) -> Dict[str, str]:
         """
         Create a new context from a message.
 
         """
-        if isinstance(context, SQSMessage):
-            return self.from_sqs_message(context, **kwargs)
-
-        # XXX we want to remove this
```
jessemyers: No longer needed

microcosm_pubsub/dispatcher.py
```diff
@@ -49,7 +49,11 @@ def handle_message(self, message, bound_handlers) -> MessageHandlingResult:
                     message=message,
                     bound_handlers=bound_handlers,
                 )
+            # save elapsed time
             message_handling_result.elapsed_time = self.opaque["elapsed_time"]
+            # log handler outcome
```
jessemyers: We now log after creating the message result.

There is still some awkwardness with this logic. It'll need one more pass I think.

microcosm_pubsub/dispatcher.py
```diff
@@ -107,60 +106,24 @@ def find_handler(self, message, bound_handlers):
         """
         try:
             handler = self.sqs_message_handler_registry.find(message.media_type, bound_handlers)
+            # XXX awkward
```
jessemyers: The "logging" function was also wrapping the handler with a decorator that ensures that all loggers have access to `extra` via a function callback. This callback used to call the `sqs_message._context` -- for no good reason since it's always performed inside of the opaque initializer context.

This code needs more testing, but it:
 - Saves the original handler (better abstraction needed)
 - Wraps automatically

microcosm_pubsub/message.py
```diff
@@ -17,14 +17,16 @@ def __init__(self,
...
+                 approximate_receive_count=None,
+                 handler=None):
         self.consumer = consumer
         self.content = content
         self.media_type = media_type
         self.message_id = message_id
         self.receipt_handle = receipt_handle
         self.topic_arn = topic_arn
         self.approximate_receive_count = approximate_receive_count
+        self.handler = handler
```
jessemyers: I am temporarily saving the "original" handler on the message. I don't intend to keep this.

microcosm_pubsub/result.py
```diff
@@ -3,68 +3,106 @@
...
+from logging import DEBUG, INFO, WARNING
+from sys import exc_info
+from typing import Any, Dict, Tuple
 
 from microcosm_pubsub.errors import IgnoreMessage, Nack, SkipMessage, TTLExpired
 
 
-@unique
+@dataclass
+class MessageHandlingResultTypeInfo:
```
jessemyers: Each message handling result knows how it should be logged.

microcosm_pubsub/result.py
```diff
@@ -3,68 +3,106 @@
...
             message.nack()
 
         return cls(
+            exc_info=exc_info_,
+            extra=extra,
             media_type=message.media_type,
             result=result,
         )
+
+    def log(self, logger, opaque):
```
jessemyers: Generic result logging.

microcosm_pubsub/result.py
```diff
@@ -3,68 +3,106 @@
...
+        return self.value.level
+
+    @property
+    def exc_info(self):
+        return self.value.exc_info
+
 
 @dataclass
 class MessageHandlingResult:
+    exc_info: Tuple[Any, Any, Any]
```
pelegdvir: nit: different var name? (`except_info`)

microcosm_pubsub/dispatcher.py
```diff
@@ -64,13 +68,8 @@ def _handle_message(self, message, bound_handlers) -> bool:
...
 
-        # we might want to handle the message
-        result = self.invoke_handler(handler, message.media_type, content)
-
-        self.logger.info(
-            "Handled {handler}",
-            extra=self.opaque.as_dict(),
-        )
+        # handle the message
+        result = handler(content)
```
pelegdvir: nice

microcosm_pubsub/dispatcher.py
jessemyers: I've now tested this locally.

microcosm_pubsub/dispatcher.py
```diff
@@ -107,60 +106,24 @@ def find_handler(self, message, bound_handlers):
         """
         try:
             handler = self.sqs_message_handler_registry.find(message.media_type, bound_handlers)
+            # XXX awkward
+            message.handler = handler
             self.opaque["handler"] = titleize(handler.__class__.__name__)
-            return handler
+            return context_logger(
```
pelegdvir: Something here seems strange. I know it's `microcosm-logging` but it would be nicer to return:
context_logger(self.opaque, handler)`

jessemyers: I agree, but I don't want to rewrite microcosm logging today.

microcosm_pubsub/dispatcher.py
```diff
@@ -107,60 +106,24 @@ def find_handler(self, message, bound_handlers):
...
+                lambda *args, **kwargs: self.opaque,
+                handler,
+                parent=handler,
+            )
         except KeyError:
             raise IgnoreMessage(f"No handler was registered for: {message.media_type}")
 
-    def invoke_handler(self, handler, media_type, content):
+    def choose_logger(self, handler):
```
pelegdvir: really nice

PR #130, opened 2018-10-17
Make result invocation clearer

jessemyers:
Changes:
 - Separate result factories for `success` (bool) and `error` (Exception)
 - Separate the pre-handler error handling from the handler invocation
 - Move wrapped handler call inwards

### DIFF
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -40,37 +40,29 @@ class SQSMessageDispatcher:
         Handle a message.
 
         """
-        # initialize a context for the current message
         with self.opaque.initialize(self.sqs_message_context, message):
-            # save elapsed time information to the context
+            handler = None
             with elapsed_time(self.opaque):
-                message_handling_result = MessageHandlingResult.invoke(
-                    func=self._handle_message,
-                    message=message,
-                    bound_handlers=bound_handlers,
-                )
-            # save elapsed time
-            message_handling_result.elapsed_time = self.opaque["elapsed_time"]
-            # log handler outcome
-            logger = self.choose_logger(message.handler)
-            message_handling_result.log(logger, self.opaque)
-        return message_handling_result
+                try:
+                    self.validate_ttl()
+                    self.validate_content(message)
+                    handler = self.find_handler(message, bound_handlers)
+                    instance = MessageHandlingResult.invoke(
+                        handler=self.wrap_handler(handler),
+                        message=message,
+                    )
+                except Exception as error:
+                    instance = MessageHandlingResult.from_error(
+                        message=message,
+                        error=error,
+                    )
 
-    def _handle_message(self, message, bound_handlers) -> bool:
-        """
-        Handle a single message.
-
-        :raises: Nack, IgnoreMessage, SkipMessage, TTLExpired
-
-        """
-        # we might just want to ignore the message
-        self.validate_ttl()
-        content = self.validate_content(message)
-        handler = self.find_handler(message, bound_handlers)
-
-        # handle the message
-        result = handler(content)
-        return bool(result)
+            instance.elapsed_time = self.opaque["elapsed_time"]
+            instance.log(
+                logger=self.choose_logger(handler),
+                opaque=self.opaque,
+            )
+            return instance
 
     def validate_ttl(self):
         """
@@ -97,8 +89,6 @@ class SQSMessageDispatcher:
         if message.content is None:
             raise IgnoreMessage(f"Could not parse message for: {message.media_type}")
 
-        return message.content
-
     def find_handler(self, message, bound_handlers):
         """
         Find the handler for a message.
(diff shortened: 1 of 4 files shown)


microcosm_pubsub/dispatcher.py
```diff
@@ -40,37 +40,29 @@ def handle_message(self, message, bound_handlers) -> MessageHandlingResult:
...
+                try:
+                    self.validate_ttl()
+                    self.validate_content(message)
+                    handler = self.find_handler(message, bound_handlers)
+                    instance = MessageHandlingResult.invoke(
+                        handler=self.wrap_handler(handler),
+                        message=message,
+                    )
+                except Exception as error:
+                    instance = MessageHandlingResult.from_error(
```
jessemyers: We handle errors from validation and handler finding separately so that `MessageHandlingResult.invoke` doesn't have to handle two different cases.

microcosm_pubsub/dispatcher.py
```diff
@@ -40,37 +40,29 @@ def handle_message(self, message, bound_handlers) -> MessageHandlingResult:
...
-
-        # handle the message
-        result = handler(content)
-        return bool(result)
+                try:
+                    self.validate_ttl()
+                    self.validate_content(message)
+                    handler = self.find_handler(message, bound_handlers)
+                    instance = MessageHandlingResult.invoke(
+                        handler=self.wrap_handler(handler),
```
jessemyers: We move the handler wrapping here so that the reference on L49 points to the actual handler (and not the wrapped one).

microcosm_pubsub/result.py
```diff
@@ -24,13 +26,13 @@ class MessageHandlingResultType(Enum):
...
     # Messaging handling failed.
-    FAILED = MessageHandlingResultTypeInfo(name="FAILED", level=WARNING, exc_info=True)
+    FAILED = MessageHandlingResultTypeInfo(name="FAILED", level=WARNING, exc_info=True, retry=True)
 
     # Message was not handled.
     IGNORED = MessageHandlingResultTypeInfo(name="IGNORED", level=DEBUG)
 
     # Messaging handling was intentionally retried.
-    RETRIED = MessageHandlingResultTypeInfo(name="RETRIED", level=INFO)
+    RETRIED = MessageHandlingResultTypeInfo(name="RETRIED", level=INFO, retry=True)
```
pelegdvir: Now we have a reason the rename `RETRIED` (ping @GabeLambda :))

microcosm_pubsub/result.py
```diff
@@ -42,69 +44,102 @@ class MessageHandlingResultType(Enum):
...
-    def invoke(cls, func, message, **kwargs):
-        exc_info_ = None
-        extra = dict()
+    def invoke(cls, handler, message: SQSMessage):
         try:
-            success = func(message=message, **kwargs)
-            if success:
-                result = MessageHandlingResultType.SUCCEEDED
+            success = handler(message.content)
+            result = cls.from_result(message, bool(success))
```
pelegdvir: I don't know if I like the explicit casting (I know - hints - but still)

microcosm_pubsub/result.py
```diff
@@ -42,69 +44,102 @@ class MessageHandlingResultType(Enum):
...
     def log(self, logger, opaque):
-        opaque.update(self.extra)
+        """
+        Log this result.
 
-        message = f"Result for media type: {self.media_type} was : {self.result} "
+        """
+        entry = f"Result for media type: {self.media_type} was : {self.result} "
+
+        opaque.update(self.extra)
```
pelegdvir: Can we replace this line with:
`extra = dict(**self.extra, **opaque.as_dict())`?

microcosm_pubsub/result.py
jessemyers: I agree, but is it worse to:

microcosm_pubsub/result.py
jessemyers: Sure.

microcosm_pubsub/result.py
jessemyers: I could also rename `retry` to something else...

microcosm_pubsub/result.py
pelegdvir: b

microcosm_pubsub/result.py
pelegdvir: Like `should_nack=True`?
I think we should keep `retry` - we do use the `retry` terminology  (`retry_timeout_seconds`, not `nack_timeout_seconds`)

GabeLambda: It's either that, or go back to using the nack terminology (i.e. NACKED).

microcosm_pubsub/result.py
GabeLambda: You can also declare `handler` as `Callable[[dict], bool]` ([doc](https://docs.python.org/3/library/typing.html#typing.Callable)), and `success: bool = [...]`. Then you shouldn't need to cast it explicitly.

jessemyers: But the handler contract isn't actually meant to return `bool`, just something truthy

GabeLambda: Really? Most of the handlers I had seen returned True/False explicitly.

PR #131, opened 2018-10-17
GLOB-26005: demonic metrics

moshez:
## Why?

We want to have visibility into the behavior of `pubsub` daemons. This has the daemons send latencies (appropriately tagged) to a `statsd` daemon using `microcosm-metrics`. Note that we make sure to include the `result` and the `media-type` as tags, enabling us to cut and slice these things at the aggregator.

## Example StatsD output

This shows how the data looks when using the Prometheus's project `statsd_exporter`.
If using Prometheus to collect the metrics,
this correlates to how things will look when using Prometheus queries.

```
# HELP message Metric autogenerated by statsd_exporter.
# TYPE message summary
message{environment="undefined",media_type="application/vnd.microcosm.derived",result="EXPIRED",service="example",source="micrcocosm-pubsub",quantile="0.5"} 2.4080276489257812e-05
message{environment="undefined",media_type="application/vnd.microcosm.derived",result="EXPIRED",service="example",source="micrcocosm-pubsub",quantile="0.9"} 2.4080276489257812e-05
message{environment="undefined",media_type="application/vnd.microcosm.derived",result="EXPIRED",service="example",source="micrcocosm-pubsub",quantile="0.99"} 2.4080276489257812e-05
message_sum{environment="undefined",media_type="application/vnd.microcosm.derived",result="EXPIRED",service="example",source="micrcocosm-pubsub"} 2.4080276489257812e-05
message_count{environment="undefined",media_type="application/vnd.microcosm.derived",result="EXPIRED",service="example",source="micrcocosm-pubsub"} 1
message{environment="undefined",media_type="application/vnd.microcosm.derived",result="SUCCEEDED",service="example",source="micrcocosm-pubsub",quantile="0.5"} 0.001641988754272461
message{environment="undefined",media_type="application/vnd.microcosm.derived",result="SUCCEEDED",service="example",source="micrcocosm-pubsub",quantile="0.9"} 0.001641988754272461
message{environment="undefined",media_type="application/vnd.microcosm.derived",result="SUCCEEDED",service="example",source="micrcocosm-pubsub",quantile="0.99"} 0.001641988754272461
message_sum{environment="undefined",media_type="application/vnd.microcosm.derived",result="SUCCEEDED",service="example",source="micrcocosm-pubsub"} 0.001641988754272461
message_count{environment="undefined",media_type="application/vnd.microcosm.derived",result="SUCCEEDED",service="example",source="micrcocosm-pubsub"} 1
```

### DIFF
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -24,6 +24,7 @@ class SQSMessageDispatcher:
         self.sqs_consumer = graph.sqs_consumer
         self.sqs_message_context = graph.sqs_message_context
         self.sqs_message_handler_registry = graph.sqs_message_handler_registry
+        self.send_metrics = graph.pubsub_send_metrics
 
     def handle_batch(self, bound_handlers) -> List[MessageHandlingResult]:
         """
@@ -62,6 +63,7 @@ class SQSMessageDispatcher:
                 logger=self.choose_logger(handler),
                 opaque=self.opaque,
             )
+            self.send_metrics(instance)
             return instance
 
     def validate_ttl(self):
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -31,10 +31,13 @@ class SQSMessageDispatcher:
         Send a batch of messages to a function.
 
         """
-        return [
+        instances = [
             self.handle_message(message, bound_handlers)
             for message in self.sqs_consumer.consume()
         ]
+        for instance in instances:
+            self.send_metrics(instance)
+        return instances
 
     def handle_message(self, message, bound_handlers) -> MessageHandlingResult:
         """
@@ -63,7 +66,6 @@ class SQSMessageDispatcher:
                 logger=self.choose_logger(handler),
                 opaque=self.opaque,
             )
-            self.send_metrics(instance)
             return instance
 
     def validate_ttl(self):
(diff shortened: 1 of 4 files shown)


microcosm_pubsub/dispatcher.py
```diff
@@ -62,6 +63,7 @@ def handle_message(self, message, bound_handlers) -> MessageHandlingResult:
                 logger=self.choose_logger(handler),
                 opaque=self.opaque,
             )
+            self.send_metrics(instance)
```
jessemyers: I think we probably want this to be invoked one or two levels higher up.

The "clock" starts for an SQS message when we read it (as part of a batch). The clock finishes when we `ack` (delete) or `nack` (extend the timer) after each message. This happens in the logic above, but only for one message.

This means we're delaying the ack/nack call for subsequent messages in a batch on earlier messages writing their metrics.

It'd be better to do so after the entire batch has been processed.

microcosm_pubsub/metrics.py
```diff
@@ -0,0 +1,34 @@
...
+from microcosm.api import defaults
+from microcosm_pubsub.result import MessageHandlingResultType
+
+
+def do_nothing(results):
+    return
+
+
+@defaults(
+    enable="false",
```
jessemyers: Use: ```
@defaults(
    enabled=typed(boolean, default_value=False)
)
```

PR #132, opened 2018-10-17
Properly ack on ignore

jessemyers:


microcosm_pubsub/dispatcher.py
```diff
@@ -62,6 +62,7 @@ def handle_message(self, message, bound_handlers) -> MessageHandlingResult:
                 logger=self.choose_logger(handler),
                 opaque=self.opaque,
             )
+            instance.resolve(message)
```
jessemyers: It was a mistake to put the `ack/nack` calls automatically in the `invoke` call because validation failures would bypass this.

microcosm_pubsub/tests/test_result.py
```diff
@@ -40,7 +40,7 @@ def handler(message):
         result = MessageHandlingResult.invoke(
             handler=handler,
             message=self.message,
-        )
+        ).resolve(self.message)
```
jessemyers: Rather to rewrite tests to validate these cases separately, I made the minimal change.

PR #139, opened 2018-10-23
Add ability to skip a message seen too many times

afallou:
LocalStack doesn't have deadlettering capabilities. To avoid processing the same failing message again and again, we make it possible to just skip it after a certain number of attempts.
- Change default backoff policy to return negative timeout after too many attempts
- Default config to max of `-1` attempts, interpreted as turning off the limit in number of attempts.

### DIFF
--- a/microcosm_pubsub/consumer.py
+++ b/microcosm_pubsub/consumer.py
@@ -102,22 +102,11 @@ class SQSConsumer:
 
         """
         timeout = self.backoff_policy.compute_backoff_timeout(message, visibility_timeout_seconds)
-        if timeout < 0:
-            return self.skip_nack(message)
         self.sqs_client.change_message_visibility(
             QueueUrl=self.sqs_queue_url,
             ReceiptHandle=message.receipt_handle,
             VisibilityTimeout=timeout,
         )
-
-    def skip_nack(self, message):
-        self.logger.info(
-            "Message has exceeded the maximum of {attempts} processing attempts. Skipping",
-            extra=dict(
-                attempts=self.backoff_policy.max_processing_attempts,
-            )
-        )
-        self.ack(message)
 
 
 def configure_sqs_client(graph):
@@ -144,9 +133,6 @@ def configure_sqs_client(graph):
     wait_seconds=typed(int, default_value=1),
     # On error, change the visibility timeout when nacking
     message_retry_visibility_timeout_seconds=typed(int, default_value=5),
-    # Number of failed attempts after which the message stops being processed
-    # Only used by the CutoffBackoffPolicy; only meant to be used in non-SQS contexts
-    message_max_processing_attempts=typed(int, default_value=-1)
 )
 def configure_sqs_consumer(graph):
     """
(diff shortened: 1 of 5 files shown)


microcosm_pubsub/consumer.py
```diff
@@ -100,12 +102,23 @@ def nack(self, message, visibility_timeout_seconds=None):
 
         """
         timeout = self.backoff_policy.compute_backoff_timeout(message, visibility_timeout_seconds)
+        if timeout < 0:
+            return self.skip_nack(message)
```
afallou: Raising a `SkipMessage` at this point in the flow doesn't work, because it's not caught. Instead we have to `ack` the message.

jessemyers: Can we change the behavior somehow so `SkipMessage` works?

afallou:
@jessemyers to be able to use `SkipMessage` I moved the skipping logic earlier. It seems more natural that way - the skipping logic decides whether we should process the message, while the backoff logic decides what to do on failed processing; it wasn't very natural to mix both.

microcosm_pubsub/dispatcher.py
```diff
@@ -81,6 +91,13 @@ def validate_ttl(self):
         if int(self.opaque[TTL_KEY]) <= 0:
             raise TTLExpired()
 
+    def validate_processing_limit(self, message):
+        if self.max_processing_attempts > 0 and message.approximate_receive_count > self.max_processing_attempts:
+            raise SkipMessage(
+                "Message exceeded maximum of {max} processing attempts. Skipping",
+                extra=dict(max=self.max_processing_attempts),
+            )
```
afallou: This logic is _very_ similar to `validate_ttl` just above.

I was going to switch and use `X-Request-Ttl` to achieve this effect. However, my understanding is that once enqueued both the body and attributes (metadata) of an SQS message are immutable.
This means that to update `X-Request-Ttl` on the message we'd have to first delete it from the queue, then re-enqueue it with the updated attributes. That would be quite a change from the way we do things today (and it feels wrong).

Because of that:
- I went with the approximate receive count
- I'm wondering if `X-Request-Ttl` is usable at all - it is if the intent is for the value to be updated by a message producer (i.e. either a gateway or the service producing the message, when re-processing the same request); it's not if the intent is for it to be updated by a daemon.

jessemyers: I think it's reasonable to split the cases; we're using TTL to "hops" a request id follows globally, not just in pubsub.

In your case, you are only looking at local pubsub behavior.

microcosm_pubsub/dispatcher.py
```diff
@@ -5,15 +5,20 @@
...
 from microcosm_pubsub.context import TTL_KEY
-from microcosm_pubsub.errors import IgnoreMessage, TTLExpired
+from microcosm_pubsub.errors import IgnoreMessage, TTLExpired, SkipMessage
 from microcosm_pubsub.result import MessageHandlingResult
 
 
 @logger
+@defaults(
+    # Number of failed attempts after which the message stops being processed
+    message_max_processing_attempts=typed(int, default_value=-1)
```
jessemyers: You could probably use `None` as the default value instead of `-1`

PR #140, opened 2018-10-26
Release 1.16.0

afallou:
Adds ability to skip a message processed too many times.

### DIFF
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -5,15 +5,20 @@ Process batches of messages.
 from typing import List
 
 from inflection import titleize
+from microcosm.api import defaults, typed
 from microcosm_logging.decorators import context_logger, logger
 from microcosm_logging.timing import elapsed_time
 
 from microcosm_pubsub.context import TTL_KEY
-from microcosm_pubsub.errors import IgnoreMessage, TTLExpired
+from microcosm_pubsub.errors import IgnoreMessage, TTLExpired, SkipMessage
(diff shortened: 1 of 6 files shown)


microcosm_pubsub/dispatcher.py
```diff
@@ -81,6 +91,13 @@ def validate_ttl(self):
         if int(self.opaque[TTL_KEY]) <= 0:
             raise TTLExpired()
 
+    def validate_processing_limit(self, message):
+        if self.max_processing_attempts and message.approximate_receive_count > self.max_processing_attempts:
+            raise SkipMessage(
+                "Message exceeded maximum of {max} processing attempts. Skipping",
+                extra=dict(max=self.max_processing_attempts),
+            )
+
```
wheremiah: If the uri is not automagically included here, can you add it?

wheremiah:
This sounds like a interesting feature.  Do you have some specific use cases in mind where this is desired, instead of letting the message fail to the DLQ?

pelegdvir:
@wheremiah -
LocalStack doesn't have deadlettering capabilities. To avoid processing the same failing message again and again, we make it possible to just skip it after a certain number of attempts.
(see: https://github.com/globality-corp/microcosm-pubsub/pull/139)

wheremiah:
@pelegdvir 
Hm, but this change applies to the whole platform... I mean I hate DLQs as much as the next guy, maybe more, but this will hide things that should have DLQ'd as SkipMessages.  That seems... risky to me.

jessemyers:
@wheremiah IIUC, the new feature is default disabled (see `@defaults`)

PR #142, opened 2018-11-06
Slice deferred message batches

GabeLambda:
AWS appears to have some limitations regarding message sizes, such that
we run into errors when trying to exceed those limits (for example, a
message containing a payload of 8MB was rejected).

This adds an extra configuration and behavior to split incoming batched
messages into smaller batches before being publishing.

### DIFF
--- a/microcosm_pubsub/producer.py
+++ b/microcosm_pubsub/producer.py
@@ -7,7 +7,7 @@ from distutils.util import strtobool
 from functools import wraps
 
 from boto3 import Session
-from microcosm.api import defaults
+from microcosm.api import defaults, typed
 from microcosm.errors import NotBoundError
 from microcosm_logging.decorators import logger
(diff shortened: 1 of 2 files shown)


microcosm_pubsub/producer.py
```diff
@@ -249,6 +256,8 @@ def configure_sns_topic_arns(graph):
     endpoint_url=None,
     mock_sns=True,
     skip=None,
+    # the size used to determine batching in the deferred batch producer
+    deferred_batch_size=typed(int, default_value=100),
```
GabeLambda: A bit awkward, but since the `DeferredBatchProducer` isn't a `graph`'d component, it seems harder to configure via this mechanism.

sethisernhagen:
Thanks @GabeLambda!! I just noticed (and ignored) these errors in ariza.

PR #143, opened 2018-11-06
Release: Slice deferred message batches

GabeLambda:
AWS appears to have some limitations regarding message sizes, such that we
run into errors when trying to exceed those limits (for example, a message
containing a payload of 8MB was rejected).

This adds an extra configuration and behavior to split incoming batched
messages into smaller batches before being publishing.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.16.0
+current_version = 1.17.0
 commit = False
 tag = False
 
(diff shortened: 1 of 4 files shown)


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.15.0
+current_version = 1.17.0
```
GabeLambda: The last version appeared to have been missed on master merge.

PR #144, opened 2018-11-15
Add matcher for published messages

jessemyers:
Extend and wrap base `hamcrest` matchers for `contains` and `hasproperty` to work
with mocked SNS producer calls.

### DIFF
--- a/microcosm_pubsub/chain/context_decorators.py
+++ b/microcosm_pubsub/chain/context_decorators.py
@@ -1,5 +1,5 @@
 from functools import wraps, WRAPPER_ASSIGNMENTS
-from inspect import signature, _POSITIONAL_ONLY, _POSITIONAL_OR_KEYWORD, _empty
+from inspect import Parameter, signature, Signature
 
 from microcosm_pubsub.chain.decorators import EXTRACTS, BINDS
 
@@ -19,7 +19,7 @@ def get_positional_args(func):
         (arg, parameter.default)
         for arg, parameter
         in signature(func).parameters.items()
-        if parameter.kind in (_POSITIONAL_ONLY, _POSITIONAL_OR_KEYWORD)
+        if parameter.kind in (Parameter.POSITIONAL_ONLY, Parameter.POSITIONAL_OR_KEYWORD)
     ]
 
 
@@ -34,7 +34,7 @@ def get_from_context(context, func, assigned=DEFAULT_ASSIGNED):
     @wraps(func, assigned=assigned + WRAPPER_ASSIGNMENTS)
     def decorate(*args, **kwargs):
         context_kwargs = {
-            arg_name: (context[arg_name] if default is _empty else context.get(arg_name, default))
+            arg_name: (context[arg_name] if default is Signature.empty else context.get(arg_name, default))
             for arg_name, default in positional_args[len(args):]
             if arg_name not in kwargs
         }
--- a/microcosm_pubsub/envelope.py
+++ b/microcosm_pubsub/envelope.py
@@ -120,7 +120,7 @@ class SQSEnvelope(MessageBodyParser, MediaTypeAndContentParser):
 
     """
     def __init__(self, graph):
-        self.validate_md5 = graph.config.sqs_envelope.validate_md5
+        self.should_validate_md5 = graph.config.sqs_envelope.validate_md5
 
     def parse_raw_message(self, consumer, raw_message):
         """
@@ -134,7 +134,7 @@ class SQSEnvelope(MessageBodyParser, MediaTypeAndContentParser):
 
         body = self.parse_body(raw_message)
 
-        if self.validate_md5:
+        if self.should_validate_md5:
             self.validate_md5(raw_message, body)
 
         message = self.parse_message(body)
--- a/microcosm_pubsub/tests/test_matcher.py
+++ b/microcosm_pubsub/tests/test_matcher.py
@@ -1,0 +1,79 @@
+"""
+Test customized matcher.
+
+"""
+from hamcrest import assert_that, all_of
+from microcosm.api import create_object_graph, load_from_dict
+
+from microcosm_pubsub.conventions import created
+from microcosm_pubsub.matchers import (
+    has_media_type,
+    has_uri,
+    published,
+    published_inanyorder,
+    published_nothing,
+)
+
+
+class TestPublishingMatcher:
+
+    def setup(self):
+        loader = load_from_dict(
+            sns_topic_arns=dict(
+                default="topic",
+            )
+        )
+        self.graph = create_object_graph("example", testing=True, loader=loader)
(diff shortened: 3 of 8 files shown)


microcosm_pubsub/chain/context_decorators.py
```diff
@@ -1,5 +1,5 @@
 from functools import wraps, WRAPPER_ASSIGNMENTS
-from inspect import signature, _POSITIONAL_ONLY, _POSITIONAL_OR_KEYWORD, _empty
+from inspect import Parameter, signature, Signature
```
jessemyers: I ran `mypy` against `microcosm-pubsub` and found a few breakages here (maybe related to Python3?)

microcosm_pubsub/envelope.py
```diff
@@ -120,7 +120,7 @@ class SQSEnvelope(MessageBodyParser, MediaTypeAndContentParser):
 
     """
     def __init__(self, graph):
-        self.validate_md5 = graph.config.sqs_envelope.validate_md5
+        self.should_validate_md5 = graph.config.sqs_envelope.validate_md5
```
jessemyers: Another `mypy` violation (name shadowing)

microcosm_pubsub/matchers.py
```diff
@@ -0,0 +1,95 @@
+from dataclasses import dataclass
```
jessemyers: The actual feature is here.

microcosm_pubsub/matchers.py
```diff
@@ -0,0 +1,95 @@
...
+from typing import List, Optional, Mapping
+
+from hamcrest import not_none
+from hamcrest.library.collection.issequence_containinginorder import IsSequenceContainingInOrder
+from hamcrest.library.object.hasproperty import IsObjectWithProperty
+from hamcrest.core.helpers.wrap_matcher import wrap_matcher
+
+
+@dataclass
+class PublishedMessage:
```
jessemyers: Having an object to encapsulate the `mock_call` allows us to use existing object matchers with the internal structure (and abstract out the JSON decoding).

microcosm_pubsub/matchers.py
```diff
@@ -0,0 +1,95 @@
...
+        return self.message.get("mediaType")
+
+    @property
+    def uri(self) -> Optional[str]:
+        return self.message.get("uri")
+
+    @classmethod
+    def from_call(cls, call) -> "PublishedMessage":
+        name, args, kwargs = call
+        topic = kwargs["TopicArn"]
```
jessemyers: The right thing to do here is to leverage the enveloping code. I'm not sure if we need to since we are very unlikely to run a unit test with these matchers and also change the default enveloping.

microcosm_pubsub/matchers.py
```diff
@@ -0,0 +1,95 @@
...
+    def from_mock_calls(sns_producer) -> List["PublishedMessage"]:
+        return [
+            PublishedMessage.from_call(mock_call)
+            for mock_call in sns_producer.sns_client.publish.mock_calls
+            # NB: _Call is a tuple of (name, args, kwargs)
+            if "TopicArn" in mock_call[-1] and "Message" in mock_call[-1]
+        ]
+
+
+class PublishedMessageMatcher(IsObjectWithProperty):
```
jessemyers: This class is purely cosmetic; it essential changes the default object matcher description to use "message" instead of "object".

This may not be worthwhile complexity.

microcosm_pubsub/matchers.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    def describe_to(self, description):
+        return description.append_text(
+            f"a message with a {self.property_name} matching ",
+        ).append_description_of(
+            self.value_matcher,
+        )
+
+
+class PublishingMatcher(IsSequenceContainingInOrder):
```
jessemyers: The main value we're adding is L68: delegate to the superclass (which is the implementation of `contains`), but translate the (mocked) SNSProducer to a a list of `PublishedMessage` object.

microcosm_pubsub/matchers.py
```diff
@@ -0,0 +1,95 @@
...
+
+    """
+    def matches(self, sns_producer, mismatch_description=None):
+        return super().matches(
+            PublishedMessage.from_mock_calls(sns_producer),
+            mismatch_description,
+        )
+
+
+def published(*matchers):
```
jessemyers: Syntactic sugar from here on out.

microcosm_pubsub/tests/test_matcher.py
```diff
@@ -0,0 +1,55 @@
...
+    def setup(self):
+        loader = load_from_dict(
+            sns_topic_arns=dict(
+                default="topic",
+            )
+        )
+        self.graph = create_object_graph("example", testing=True, loader=loader)
+        self.graph.sns_producer.sns_client.reset_mocks()
+
+    def test_publish_no_messages(self):
```
jessemyers: Examples here.

microcosm_pubsub/matchers.py
```diff
@@ -0,0 +1,95 @@
...
+def has_media_type(match=None):
+    if match is None:
+        match = not_none()
+
+    return PublishedMessageMatcher("media_type", wrap_matcher(match))
+
+
+def has_uri(match=None):
+    if match is None:
+        match = not_none()
```
pelegdvir: funny

jessemyers: I didn't think about how this would read. Oddly, it is the expected semantics: if you want to match against `published(has_uri())`, we care that the message *has-a* `uri`, but nothing else.

microcosm_pubsub/matchers.py
pelegdvir: It's nice

microcosm_pubsub/matchers.py
```diff
@@ -0,0 +1,95 @@
...
+        return cls(
+            topic=topic,
+            message=message,
+        )
+
+    @staticmethod
+    def from_mock_calls(sns_producer) -> List["PublishedMessage"]:
+        return [
+            PublishedMessage.from_call(mock_call)
+            for mock_call in sns_producer.sns_client.publish.mock_calls
```
pelegdvir: Adding a reference to `sns_client.publish.mock_calls` can be helpful

microcosm_pubsub/matchers.py
pelegdvir: I agree it's the expected semantics and the right thing to do - just odd

microcosm_pubsub/chain/context_decorators.py
GabeLambda: What were the breakages?

GabeLambda: Ah, I see you're calling it off `Parameter`. I think `mypy` has troubles validating some kinds of otherwise valid imports.

PR #149, opened 2019-02-08
Add message reason to extra dictionary

saivickna:
Why?

When we have multiple SkipMessages and provide a reason, it will not show up in the log.


microcosm_pubsub/result.py
```diff
@@ -2,13 +2,14 @@
...
-from dataclasses import dataclass, field
 from enum import Enum, unique
 from logging import DEBUG, INFO, WARNING
 from sys import exc_info
 from typing import Any, Dict, Optional, Tuple
 
-from microcosm_pubsub.errors import IgnoreMessage, Nack, SkipMessage, TTLExpired
+from dataclasses import dataclass, field
+from microcosm_pubsub.errors import (IgnoreMessage, Nack, SkipMessage,
+                                     TTLExpired)
```
GabeLambda: Nit: wrong import breaks.

microcosm_pubsub/result.py
```diff
@@ -2,13 +2,14 @@
...
 
 """
-from dataclasses import dataclass, field
 from enum import Enum, unique
 from logging import DEBUG, INFO, WARNING
 from sys import exc_info
 from typing import Any, Dict, Optional, Tuple
 
-from microcosm_pubsub.errors import IgnoreMessage, Nack, SkipMessage, TTLExpired
+from dataclasses import dataclass, field
```
GabeLambda: dataclasses should be in the stdlib in 3.7+.

microcosm_pubsub/tests/test_result.py
```diff
@@ -2,18 +2,16 @@
...
 """
-from hamcrest import (
-    assert_that,
-    has_properties,
-)
+from hamcrest import assert_that, has_entries, has_properties
 from microcosm_logging.decorators import logger
 
-from microcosm_pubsub.errors import IgnoreMessage, Nack, SkipMessage, TTLExpired
+from microcosm_pubsub.errors import (IgnoreMessage, Nack, SkipMessage,
```
GabeLambda: Nit: import linebreaks.

PR #150, opened 2019-03-22
Make switch statement rely on __eq__

afallou:
**Why?**
The switch statement was working with an underlying dictionary of cases. This prohibits having non-hashable values in either the value evaluated in `switch` or the value compared against in each `case`.

**What?**
- Use a list of tuple instead, so we can evaluate the logical branching with `__eq__` instead of `__hash__`.
- Make `TryChain` inherit from `Switch`.

This second change item is because the first one broke `TryChainStatement`, which relies on `CaseStatement`. Since `try_chain` is really just syntactic sugar around the `switch().case()`

### DIFF
--- a/microcosm_pubsub/chain/statements/try_chain.py
+++ b/microcosm_pubsub/chain/statements/try_chain.py
@@ -12,30 +12,22 @@ try(
 """
 from microcosm_pubsub.chain import Chain
 from microcosm_pubsub.chain.statements.case import CaseStatement
+from microcosm_pubsub.chain.statements.switch import SwitchStatement
 
 
-class TryChainStatement:
+class TryChainStatement(SwitchStatement):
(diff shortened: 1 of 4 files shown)


microcosm_pubsub/chain/statements/try_chain.py
```diff
@@ -12,30 +12,22 @@
...
         self.chain = Chain.make(*args, **kwargs)
-        self._cases = dict()
-        self._otherwise = None
 
     def catch(self, key, *args,  **kwargs):
-        if not args and not kwargs:
-            return CaseStatement(self, key)
-
-        self._cases[key] = Chain.make(*args, **kwargs)
-        return self
```
afallou: This is exactly the same as `Switch`'s `case`.

PR #157, opened 2019-05-13
Add metrics per-handled message batch

GabeLambda:
Adds two new metrics: the time needed to process a batch of messages,
and the count of relevant messages per batch.

The previous metrics, while useful for per-message handling statistics,
weren't useful enough when trying to analyze message-handling from the
consumer's viewpoint. For example, it would be difficult to see how
long, under load, it took for any given service to handle a batch of
messages.

### DIFF
--- a/microcosm_pubsub/metrics.py
+++ b/microcosm_pubsub/metrics.py
@@ -1,6 +1,7 @@
 from microcosm.api import defaults, typed
-from microcosm.errors import NotBoundError
 from microcosm.config.types import boolean
+from microcosm.errors import NotBoundError
+
 from microcosm_pubsub.result import MessageHandlingResultType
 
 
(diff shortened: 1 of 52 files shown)


microcosm_pubsub/metrics.py
```diff
@@ -50,3 +54,53 @@ def __call__(self, result):
...
+
+        tags = [
+            "source:microcosm-pubsub",
+        ]
+
+        self.metrics.histogram(
+            "message_batch",
+            elapsed_time,
+            tags=tags,
+        )
```
GabeLambda: Recording histograms for per-batch handling time.

microcosm_pubsub/metrics.py
```diff
@@ -50,3 +54,53 @@ def __call__(self, result):
...
+
+        self.metrics.histogram(
+            "message_count",
+            message_count,
+            tags=tags,
+        )
```
GabeLambda: And the usual rate of messages handled per batch.

PR #158, opened 2019-05-13
Add metrics per-handled message batch

GabeLambda:
Adds two new metrics: the time needed to process a batch of messages, and
the count of relevant messages per batch.

The previous metrics, while useful for per-message handling statistics,
weren't useful enough when trying to analyze message-handling from the
consumer's viewpoint. For example, it would be difficult to see how long,
under load, it took for any given service to handle a batch of messages.

### DIFF
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -11,7 +11,7 @@ from microcosm_logging.timing import elapsed_time
 
 from microcosm_pubsub.context import TTL_KEY
 from microcosm_pubsub.errors import IgnoreMessage, SkipMessage, TTLExpired
-from microcosm_pubsub.result import MessageHandlingResult
+from microcosm_pubsub.result import MessageHandlingResult, MessageHandlingResultType
 
 
 @logger
@@ -30,6 +30,7 @@ class SQSMessageDispatcher:
         self.sqs_message_context = graph.sqs_message_context
         self.sqs_message_handler_registry = graph.sqs_message_handler_registry
         self.send_metrics = graph.pubsub_send_metrics
+        self.send_batch_metrics = graph.pubsub_send_batch_metrics
         self.max_processing_attempts = graph.config.sqs_message_dispatcher.message_max_processing_attempts
 
     def handle_batch(self, bound_handlers) -> List[MessageHandlingResult]:
@@ -37,12 +38,30 @@ class SQSMessageDispatcher:
         Send a batch of messages to a function.
 
         """
-        instances = [
-            self.handle_message(message, bound_handlers)
-            for message in self.sqs_consumer.consume()
-        ]
+        with elapsed_time(self.opaque):
+            instances = [
+                self.handle_message(message, bound_handlers)
+                for message in self.sqs_consumer.consume()
+            ]
+
+        message_count = len([
+            instance for instance in instances
+            if instance.result != MessageHandlingResultType.IGNORED
+        ])
+
+        self.logger.info(
+            f"Finished handling batch",
+            extra=dict(
+                message_count=message_count,
+                **self.opaque.as_dict(),
+            ),
+        )
+
+        self.send_batch_metrics(self.opaque["elapsed_time"], message_count)
+
         for instance in instances:
             self.send_metrics(instance)
+
         return instances
 
     def handle_message(self, message, bound_handlers) -> MessageHandlingResult:
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -50,7 +50,7 @@ class SQSMessageDispatcher:
         ])
 
         self.logger.info(
-            f"Finished handling batch",
+            "Finished handling batch",
             extra=dict(
                 message_count=message_count,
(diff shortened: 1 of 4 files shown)


microcosm_pubsub/dispatcher.py
```diff
@@ -30,19 +30,38 @@ def __init__(self, graph):
...
+                self.handle_message(message, bound_handlers)
+                for message in self.sqs_consumer.consume()
+            ]
+
+        message_count = len([
+            instance for instance in instances
+            if instance.result != MessageHandlingResultType.IGNORED
+        ])
+
+        self.logger.info(
```
GabeLambda: Logs and tracks message batch handling outside of a single request-id. This may be less than useful, since the logs will intermingle with different runs of a single handler. It should still be possible to narrow down by tag.

microcosm_pubsub/metrics.py
```diff
@@ -51,3 +54,53 @@ def __call__(self, result):
...
+        """
+        if not self.enabled:
+            return
+
+        tags = [
+            "source:microcosm-pubsub",
+        ]
+
+        self.metrics.histogram(
+            "message_batch",
```
GabeLambda: Time to handle batch.

microcosm_pubsub/metrics.py
```diff
@@ -51,3 +54,53 @@ def __call__(self, result):
...
+            elapsed_time,
+            tags=tags,
+        )
+
+        self.metrics.histogram(
+            "message_batch_count",
```
GabeLambda: Ends up being the approximate count per-batch of events that were actually handled (e.g. checks proportion of events out of the default 5/10 that were relevant).

microcosm_pubsub/dispatcher.py
```diff
@@ -24,25 +25,46 @@ class SQSMessageDispatcher:
...
     def handle_batch(self, bound_handlers) -> List[MessageHandlingResult]:
         """
         Send a batch of messages to a function.
 
         """
-        instances = [
-            self.handle_message(message, bound_handlers)
-            for message in self.sqs_consumer.consume()
-        ]
+        with elapsed_time(self.opaque):
```
cdm-ium: I don't think we want to put this logging in `self.opaque`. This is similar to what I was doing on `request-instances`. If you put it in `self.opaque`, it will pick up a mysterious `elapsed_time` key which has no meaning once it leaves this service.

microcosm_pubsub/dispatcher.py
```diff
@@ -24,25 +25,46 @@ class SQSMessageDispatcher:
...
+            instances = [
+                self.handle_message(message, bound_handlers)
+                for message in self.sqs_consumer.consume()
+            ]
+
+        message_count = len([
+            instance for instance in instances
+            if instance.result != MessageHandlingResultType.IGNORED
```
cdm-ium: I think knowing about ignored messages is potentially interesting too, somewhere. Just to make sure they aren't burning resources.

microcosm_pubsub/dispatcher.py
cdm-ium: Yeah, at the very least I don't think we should put the `elapsed_time` here in `self.opaque`.  For the short term I'd just use a local variable to this handler, but we can think about what we want to do more system wide as we aren't finding the similar problem elsewhere.

microcosm_pubsub/dispatcher.py
```diff
@@ -24,25 +25,46 @@ class SQSMessageDispatcher:
...
+
+        self.logger.info(
+            "Finished handling batch",
+            extra=dict(
+                message_count=message_count,
+                **self.opaque.as_dict(),
+            ),
+        )
+
+        self.send_batch_metrics(self.opaque["elapsed_time"], message_count)
```
cdm-ium: Again this feels messy because of how `self.opaque` will be propagated.

microcosm_pubsub/metrics.py
```diff
@@ -51,3 +51,53 @@ def __call__(self, result):
             result.elapsed_time,
             tags=tags,
         )
+
+
+@defaults(
+    enabled=typed(boolean, default_value=True)
+)
+class PubSubSendBatchMetrics:
```
cdm-ium: Docstring?

microcosm_pubsub/dispatcher.py
GabeLambda: Fair enough.

microcosm_pubsub/dispatcher.py
GabeLambda: Right. I'll make another object for this.

microcosm_pubsub/dispatcher.py
GabeLambda: I was thinking about that, actually. The problem then is sort of then you'd get a constant "10" here, and that's not super useful. I think it's a safe assumption (for now) that `IGNORED` messages are fast enough (since it should be a relegated to a single hash table lookup).

microcosm_pubsub/metrics.py
GabeLambda: Sure.

PR #164, opened 2019-05-16
Release 1.24.3

GabeLambda:


microcosm_pubsub/producer.py
```diff
@@ -51,6 +53,8 @@ def create_message(self, media_type, dct, uri=None, opaque_data=None, **kwargs):
         if self.opaque is not None:
             opaque_data.update(self.opaque.as_dict())
 
+        opaque_data[PUBLISHED_KEY] = time()
```
GabeLambda: Noting here that yes, this isn't 100% accurate, given that a message may be either deferred, and there is a non-zero time between message creation and message published. Unfortunately, opaque_data is encoded with the message itself, meaning that they need to be tied with each other. This could be refactored, but that is out of scope for this change.

PR #165, opened 2019-05-24
Enable message attributes + raw message parsing

GabeLambda:
Up until now, we've basically supported a simple model wherein every
daemon is notified about every event, with message filtering being done
at the consumer level. This is a tad wasteful and increases message
delivery time.

SNS supports the ability add subscription-level filters, such that they
only receive a subset of messages based on the attached metadata.

To support message filtering in our services / daemons, we need to be
able to:
* Attach message attributes to published messages
* Parse message sent via raw message delivery

For these purposes, I've initially chosen a simple media type-based
filtering. While SNS supports more complex conditions, this should be
sufficient for now.

The SNS/SQS documentation implies that raw message parsing is needed to
be able to support message filtering. While that claim is a tad dubious
(I was able to accomplish filtering without it), adding support
shouldn't hurt.

Note that this isn't quite yet ready; there is some more validation I want to do first.

### DIFF
--- a/microcosm_pubsub/codecs.py
+++ b/microcosm_pubsub/codecs.py
@@ -85,7 +85,11 @@ class PubSubMessageCodec:
         Uses the appropriate coded to read JSON.
 
         """
-        dct = loads(message)
+        try:
+            dct = loads(message)
+        except TypeError:
+            # message is already a dict
+            dct = message
         # we need to explicitly validate because dump() doesn't
         self.schema.validate(dct)
         decoded = self.schema.dump(dct)
--- a/microcosm_pubsub/codecs.py
+++ b/microcosm_pubsub/codecs.py
@@ -85,10 +85,9 @@ class PubSubMessageCodec:
         Uses the appropriate coded to read JSON.
 
         """
-        try:
+        if not isinstance(message, dict):
             dct = loads(message)
-        except TypeError:
-            # message is already a dict
+        else:
             dct = message
         # we need to explicitly validate because dump() doesn't
         self.schema.validate(dct)
--- a/microcosm_pubsub/producer.py
+++ b/microcosm_pubsub/producer.py
@@ -3,9 +3,11 @@ Message producer.
 
 """
 from collections import defaultdict
+from dataclasses import dataclass
 from distutils.util import strtobool
 from functools import wraps
 from time import time
+from typing import Dict, List
 
 from boto3 import Session
 from microcosm.api import defaults, typed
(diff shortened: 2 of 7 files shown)


microcosm_pubsub/producer.py
```diff
@@ -43,10 +59,10 @@ def produce(self, media_type, dct=None, uri=None, **kwargs):
 
         if self.skip:
             return
-        message, topic_arn, opaque_data = self.create_message(media_type, dct, uri, **kwargs)
-        return self.publish_message(media_type, message, topic_arn, opaque_data)
+        pubsub_message = self.create_message(media_type, dct, uri, **kwargs)
```
GabeLambda: This list was getting longer, so I just put all of the needed fields into a container.

microcosm_pubsub/producer.py
```diff
@@ -97,6 +123,18 @@ def choose_topic_arn(self, media_type):
             ))
         return topic_arn
 
+    def choose_message_attributes(self, media_type: str) -> Dict[str, Dict[str, str]]:
```
GabeLambda: The new attributes.

microcosm_pubsub/producer.py
```diff
@@ -132,28 +170,30 @@ def generate_message_batches(self, messages, deferred_batch_size):
...
+    def __exit__(self, type, value, traceback):
+        if type is not None:
+            return
+
+        for message_batch in self.generate_message_batches(self.messages, self.producer.deferred_batch_size):
             if len(message_batch) > 1:
                 self.producer.produce(
                     MessageBatchSchema.MEDIA_TYPE,
-                    messages=message_batch,
+                    messages=self.construct_batch_pubsub_message(message_batch)
```
GabeLambda: Had to do a bit of extra finagling here. Also need to verify that this actually works (I might break echo?)

microcosm_pubsub/codecs.py
```diff
@@ -85,7 +85,11 @@ def decode(self, message):
         Uses the appropriate coded to read JSON.
 
         """
-        dct = loads(message)
+        try:
+            dct = loads(message)
+        except TypeError:
+            # message is already a dict
```
pelegdvir: In this case, we can use `isinstance` instead.

afallou: @pelegdvir some folks would say that using `try`/`except` is more Pythonic ("ask for forgiveness" and all that)

microcosm_pubsub/producer.py
afallou: You're missing a trailing comma.

PR #170, opened 2019-06-27
Log publish messages as debug.

cdm-ium:
In terms of numbers of logs, these logs make up about 6% of our logs in
the last day and yet they provide minimal value.

### DIFF
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -62,7 +62,7 @@ class SQSMessageDispatcher:
                 message_batch_size=message_batch_size,
                 batch_elapsed_time=batch_elapsed_time,
             )
-            self.logger.info(message)
+            self.logger.debug(message)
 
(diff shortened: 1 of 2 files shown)


microcosm_pubsub/dispatcher.py
```diff
@@ -62,7 +62,7 @@ def handle_batch(self, bound_handlers) -> List[MessageHandlingResult]:
                 message_batch_size=message_batch_size,
                 batch_elapsed_time=batch_elapsed_time,
             )
-            self.logger.info(message)
+            self.logger.debug(message)
```
cdm-ium: @GabeLambda  These are the batch logs you added to get stats on sqs batch publishing times. A) I think we can ultimately catch this in jaeger B) do you really need this log right now or do you feel like you've done enough tuning with the batch sizes for now?

GabeLambda: They're debug. I'm getting additional logs for them in prometheus.

PR #171, opened 2019-06-27
Syntactic sugar to give names to function argument and result

afallou:
We don't have an easy, explicity way to temporarily change the context for a single function call.
This limits the amount of code reuse we can do: currently we'd need to write two functions with different argument names, calling the same underlying logic.
This new syntax allows us to do that much more easily.

### DIFF
--- a/microcosm_pubsub/chain/statements/call.py
+++ b/microcosm_pubsub/chain/statements/call.py
@@ -1,0 +1,51 @@
+"""
+Syntactic sugar using existing primitives defined
+
+"""
+from microcosm_pubsub.chain import Chain
+from microcosm_pubsub.chain.context import ScopedSafeContext
+
+
+class LocalCallWrapper:
+    """
+    Syntactic sugar making it easy to call the same function multiple times in a chain
+    Usage:
+    Chain(
+       ...,
+       call(a_function).with_args(a_function_arg="an_existing_context_key").as_("name_of_the_result_in_context"),
+       call(a_function).with_args(a_function_arg="some_other_key").as_("another_name"),
+    )
+
+    """
+    def __init__(self, func):
+        self.func = func
+        self.chain = Chain(func)
+        self.kwargs_mapping = dict()
+        self.result_name = None
+        self.local_context = None
+
+    def with_args(self, **kwargs):
+        self.kwargs_mapping = kwargs
+        return self
+
+    def as_(self, name):
+        self.result_name = name
+        return self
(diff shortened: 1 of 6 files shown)


microcosm_pubsub/chain/statements/call.py
```diff
@@ -0,0 +1,51 @@
+"""
+Syntactic sugar using existing primitives defined
+
+"""
+from microcosm_pubsub.chain import Chain
+from microcosm_pubsub.chain.context import ScopedSafeContext
+
+
+class LocalCallWrapper:
```
afallou: The approach taken here is somewhat similar to what we do with `for_each`. We create a sub-chain with a scoped context.
My first approach was using a combination of `binds` and `assign_function`; but `binds` doesn't allow us to do the same operation twice on the same function. Besides, the end result is much clearer this way.

microcosm_pubsub/chain/statements/call.py
```diff
@@ -0,0 +1,51 @@
...
+    def with_args(self, **kwargs):
+        self.kwargs_mapping = kwargs
+        return self
+
+    def as_(self, name):
+        self.result_name = name
+        return self
+
+    def _build_local_kwargs(self, parent_context):
+        return {
```
pelegdvir: Please use context/safe context from https://github.com/globality-corp/microcosm-pubsub/blob/develop/microcosm_pubsub/chain/context.py

afallou: @pelegdvir that's what I'm doing (see `__call__` below) - but I need the logic to pass the right kwargs to the scoped context.

pelegdvir: Right!

microcosm_pubsub/tests/chain/statements/test_call.py
```diff
@@ -0,0 +1,85 @@
...
+    def test_two_calls_of_same_function(self):
+        def double_it(input):
+            return 2 * input
+
+        def multiply(double_one, double_three):
+            return double_one * double_three
+
+        chain = Chain(
+            call(double_it).with_args(input="one").as_("double_one"),
+            call(double_it).with_args(input="three").as_("double_three"),
```
pokey: It's confusing that string literals pull from context but other literals don't.  I'd lean towards either not supporting literals or wrapping these strings in something.  Could be called `extract_from_context` or something

PR #175, opened 2019-08-19
Accommodate for marshmallow 3.0 updates

GabeLambda:
* Remove `.data` accesses
* Pass in media_type explicitly

### DIFF
--- a/microcosm_pubsub/codecs.py
+++ b/microcosm_pubsub/codecs.py
@@ -75,8 +75,7 @@ class PubSubMessageCodec:
         """
         message = dct.copy() if dct else dict()
         message.update(kwargs)
-        encoded = self.schema.load(message)
-        return dumps(encoded.data)
+        return dumps(self.schema.load(message))
 
     def decode(self, message):
         """
@@ -91,5 +90,4 @@ class PubSubMessageCodec:
             dct = message
         # we need to explicitly validate because dump() doesn't
         self.schema.validate(dct)
-        decoded = self.schema.dump(dct)
-        return decoded.data
+        return self.schema.dump(dct)
--- a/microcosm_pubsub/codecs.py
+++ b/microcosm_pubsub/codecs.py
@@ -80,6 +80,9 @@ class PubSubMessageCodec:
         """
         message = dct.copy() if dct else dict()
         message.update(kwargs)
+        # We allow unknown fields to pass through (but not included here)
+        # to accommodate unconditional parameter passing during produce()
+        # e.g. passing in `uri` for IdentityMessages
         return dumps(self.schema.load(message, unknown=EXCLUDE))
 
     def decode(self, message):
@@ -94,5 +97,6 @@ class PubSubMessageCodec:
         else:
             dct = message
         # load performs a validation and raises on error
+        # Similarly, we exclude unknown fields to avoid errors on decode
         self.schema.load(dct, unknown=EXCLUDE)
         return self.schema.dump(dct)
--- a/microcosm_pubsub/codecs.py
+++ b/microcosm_pubsub/codecs.py
@@ -4,7 +4,12 @@ Message encoding and decoding.
 """
 from json import dumps, loads
 
-from marshmallow import Schema, ValidationError, fields
+from marshmallow import (
+    EXCLUDE,
+    Schema,
+    ValidationError,
+    fields,
+)
 
 
 DEFAULT_MEDIA_TYPE = "application/json"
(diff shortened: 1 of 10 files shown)


microcosm_pubsub/producer.py
```diff
@@ -81,6 +81,7 @@ def create_message(self, media_type, dct, uri=None, opaque_data=None, **kwargs)
         message = self.pubsub_message_schema_registry.find(media_type).encode(
             dct,
             opaque_data=opaque_data,
+            media_type=media_type,
```
GabeLambda: I don't know how this was working before, but without this, it just defaults to `application/json`.

GabeLambda: Well, it defaulting now because the field's value on `missing` is the `DEFAULT_MEDIA_TYPE`, which is `application/json`. I'm still not entirely sure why the de/serialize functions aren't being called (they should be on load/dump), but it doesn't hurt to explicitly specify the media type when creating the message (we find the message by media_type, so it's a bit it should just work out the same).

microcosm_pubsub/codecs.py
```diff
@@ -75,21 +80,19 @@ def encode(self, dct=None, **kwargs):
         """
         message = dct.copy() if dct else dict()
         message.update(kwargs)
-        encoded = self.schema.load(message)
-        return dumps(encoded.data)
+        return dumps(self.schema.load(message, unknown=EXCLUDE))
```
GabeLambda: This is to accommodate during `produce` where we pass in `uri` unconditionally, which otherwise fails the `IdentityMessage` validation.

microcosm_pubsub/codecs.py
```diff
@@ -75,21 +80,19 @@ def encode(self, dct=None, **kwargs):
...
         if not isinstance(message, dict):
             dct = loads(message)
         else:
             dct = message
-        # we need to explicitly validate because dump() doesn't
-        self.schema.validate(dct)
-        decoded = self.schema.dump(dct)
-        return decoded.data
+        # load performs a validation and raises on error
+        self.schema.load(dct, unknown=EXCLUDE)
```
GabeLambda: Same comment here about discarding unknown fields.

microcosm_pubsub/fields.py
```diff
@@ -24,7 +24,7 @@ def _serialize(self, value, attr, obj):
         else:
             return value
 
-    def _deserialize(self, value, attr, data):
+    def _deserialize(self, value, attr, data, **kwargs):
```
GabeLambda: -> 3.0 migration.

microcosm_pubsub/registry.py
```diff
@@ -74,7 +73,7 @@ def find(self, media_type):
         try:
             # use a concrete schema class if any
             schema_cls = self._mappings[media_type]
-            schema = schema_cls(strict=self.strict)
+            schema = schema_cls()
```
GabeLambda: No more strict.

microcosm_pubsub/tests/conventions/test_messages.py
```diff
@@ -41,6 +41,7 @@ def test_encode_uri_message_schema():
         loads(codec.encode(
             opaque_data=dict(foo="bar"),
             uri="http://example.com",
+            media_type="application/vnd.globality.pubsub._.deleted.foo",
```
GabeLambda: Passing in media_type explicitly to en/decodes.

microcosm_pubsub/tests/test_matcher.py
```diff
@@ -37,7 +37,7 @@ def test_publish_no_messages(self):
         )
 
     def test_publish_one_message(self):
-        self.graph.sns_producer.produce(created("foo"), data="data")
+        self.graph.sns_producer.produce(created("foo"), data="data", uri="http://some-uri")
```
GabeLambda: These are registered as URIMessages; we were probably ignoring validation earlier.

microcosm_pubsub/codecs.py
pelegdvir: Please add a comment

microcosm_pubsub/tests/test_fields.py
```diff
@@ -35,21 +36,25 @@ def test_uuid_load():
         "uuid_uuid": UUID(UUID_STR),
     })
 
-    assert_that(result.data['uuid_str'], is_(equal_to(UUID_STR)))
-    assert_that(result.data['uuid_uuid'], is_(equal_to(UUID_STR)))
+    assert_that(result['uuid_str'], is_(equal_to(UUID_STR)))
+    assert_that(result['uuid_uuid'], is_(equal_to(UUID_STR)))
```
afallou: nit: double quotes

PR #177, opened 2019-08-23
Support caching of resources as part of URIHandlers and derived handlers

adamhadani:
See [This Tech Design](https://globality.atlassian.net/wiki/spaces/GLOB/pages/918424401/Tech+Design+Cache+backend+service+responses) for details.

* Implements simple memcached-based interface
* Lazy caching strategy to begin with since we do not actually have access to full resource as part of calls to the producer.produce()


microcosm_pubsub/caching/base.py
```diff
@@ -0,0 +1,21 @@
+"""
+Cache abstractions for use with PubSub resources.
+
+"""
+from abc import ABC, abstractmethod
+
+
+class CacheBase(ABC):
```
adamhadani: Just a simple uniform ABC for any implementation of resource cache..

microcosm_pubsub/caching/memcached.py
```diff
@@ -0,0 +1,88 @@
...
+
+def json_serializer(key, value):
+    """
+    Simple JSON serializer for use with caching backends
+    that only support string/bytes value storage.
+
+    Memcached is the primary use case.
+
+    """
+    if isinstance(value, str) or isinstance(value, bytes):
```
adamhadani: apparently this memcached driver does some conversion to bytes under the hood for strings, so including both here

microcosm_pubsub/caching/memcached.py
```diff
@@ -0,0 +1,88 @@
...
+        deserializer=json_deserializer,
+        testing=False,
+    ):
+        client_kwargs = dict(
+            server=(host, port),
+            serializer=json_serializer,
+            deserializer=json_deserializer,
+        )
+        if testing:
+            self.client = MockMemcacheClient(**client_kwargs)
```
adamhadani: this library gives us a nice mock client we can use with our unit-testing. nice.

microcosm_pubsub/caching/resource_cache.py
```diff
@@ -0,0 +1,27 @@
+from microcosm.api import defaults, typed
+from microcosm.config.types import boolean
+from microcosm_pubsub.caching.memcached import MemcachedCache
+
+
```
adamhadani: This is the actual microcosm entry point (defined in setup.py see below) that we can hook into with config values e.g. from ansible on a per-service level

microcosm_pubsub/handlers/uri_handler.py
```diff
@@ -54,6 +56,12 @@ def nack_if_not_found(self):
     def resource_type(self):
         return dict
 
+    def get_resource_cache(self, graph):
+        try:
+            return graph.resource_cache
+        except (LockedGraphError, NotBoundError):
```
adamhadani: I've seen NotBoundError used for similar check in `metrics.py`, however I got a `LockedGraphError`, so checking for both.

microcosm_pubsub/handlers/uri_handler.py
```diff
@@ -132,15 +140,24 @@ def get_resource(self, message, uri):
         Passes message context.
 
         """
+        if self.resource_cache:
```
adamhadani: will only evaluate to truth-y value if "resource_cache" is included in graph.use(...) *and* graph.config.resource_cache.enabled is set to `True`

setup.py
```diff
@@ -42,6 +43,7 @@
             "pubsub_lifecycle_change = microcosm_pubsub.conventions:LifecycleChange",
             "pubsub_send_batch_metrics = microcosm_pubsub.metrics:PubSubSendBatchMetrics",
             "pubsub_send_metrics = microcosm_pubsub.metrics:PubSubSendMetrics",
+            "resource_cache = microcosm_pubsub.caching:configure_resource_cache",
```
adamhadani: We may want to extract this resource caching stuff and other helpers into some `microcosm-caching` for a more standalone solution we can use also in non-daemon use cases. WDYT @GabeLambda  @pelegdvir ?

microcosm_pubsub/caching/resource_cache.py
```diff
@@ -0,0 +1,27 @@
...
+@defaults(
+    enabled=typed(boolean, default_value=True),
```
adamhadani: Might be safer to set default value to `False` before we make a release here.. WDYT?

microcosm_pubsub/caching/resource_cache.py
```diff
@@ -0,0 +1,27 @@
...
+    host="localhost",
+    port=typed(int, default_value=11211),
+)
```
pelegdvir: We want to auto-cache only some whitelisted resources - not everything (maybe we can user regex?, schema def?).
This is a good place to configure that

microcosm_pubsub/handlers/uri_handler.py
pelegdvir: and if it's a resource that we want to cache (we don't want to cache mutable objects)

setup.py
```diff
@@ -23,6 +23,7 @@
         "microcosm>=2.12.0",
         "microcosm-daemon>=1.0.0",
         "microcosm-logging>=1.3.0",
+        "pymemcache>=2.2.2",
```
GabeLambda: This is not to say that we're necessarily tied to it, but we've been using [python-memcached](https://pypi.org/project/python-memcached/) in some of our other internal services. @pelegdvir would probably know better why (iirc, there may have been some discrepancies between how it decided consistent hashing strategies with our JS strategies). This is probably not relevant to a python-only implementation, but just putting it out there.

GabeLambda: This could probably also go in under `extras` instead. You shouldn't need to worry about caching by default.

microcosm_pubsub/caching/resource_cache.py
adamhadani: @pelegdvir good catch - was meaning to do something like that e.g check for `Event` / `_event` but forgot to include - will add this in as some `whitelist_mask` parameter or so

setup.py
adamhadani: @GabeLambda good points.

Re #1 - that library seems unmaintained (last update in 2017). The one i'm using is pure python so very lightweight and seems well maintained (last updated few days ago, and maintained by pinterest). I'd be curious to give it a go and if works well can switch the other one around. WDYT?

Re #2 - crossed my mind too, I'll make this change.

microcosm_pubsub/handlers/uri_handler.py
```diff
@@ -34,6 +35,7 @@ class URIHandler:
 
     def __init__(self, graph, retry_nack_timeout=1, resource_nack_timeout=1):
         self.opaque = graph.opaque
+        self.resource_cache = self.get_resource_cache(graph)
```
GabeLambda: I was actually thinking we would subclass this instead, and override get_resource there (e.g. into a `CachingURIHandler`). I'd prefer to do that and update selectively, mostly because I want to have granular control over which resources we start to cache, instead of having it turned on for everyone just because I happened to enable the caching component.

setup.py
GabeLambda: I definitely want us to invest in a shared caching library at some point. Let's start using it in a few places first before that, though.

microcosm_pubsub/caching/resource_cache.py
GabeLambda: Selective updates preferred for big changes like this. Roll out slowly, so rollback isn't painful.

setup.py
pelegdvir: Yep. We've used the other lib for consistent JS sync (JS memcache libs are awful)

setup.py
pelegdvir: I like the idea.
Check our own `lethe` service

microcosm_pubsub/caching/resource_cache.py
GabeLambda: Which is to say, yes, false by default.

microcosm_pubsub/handlers/uri_handler.py
pelegdvir: Hmm. I prefer to avoid creating another handler.
If we want to enable only for few handlers - then fine, but as a handler property (default value is false for now)

adamhadani: I tend to agree with @pelegdvir we can control more easily with a per-instance cache_enabled property. will add something along these lines.

PR #178, opened 2019-08-24
Resource Caching Improvements

adamhadani:
Building on the previous PR #177 this PR adds a couple of additional enhancements:

* Connect and Read timeout settings for memcache client to avoid hangups on network issues
* Include media-type of pubsub message as part of context for resource whitelisting callable. This in particular is used to make sure we do not try to use the cache to get or set from the cache for pubsub messages which are not strictly 'Create' ones. This avoids any issues to do with the fact that some Event resources currently are mutable (in particular FactEvent and ChatroomEvent, for purposes of re-encryption and redaction)

### DIFF
--- a/microcosm_pubsub/caching/resource_cache.py
+++ b/microcosm_pubsub/caching/resource_cache.py
@@ -8,6 +8,8 @@ from microcosm_pubsub.caching.memcached import MemcachedCache
     enabled=typed(boolean, default_value=False),
     host="localhost",
     port=typed(int, default_value=11211),
+    connect_timeout=3,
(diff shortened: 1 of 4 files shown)


microcosm_pubsub/caching/resource_cache.py
```diff
@@ -8,6 +8,8 @@
     enabled=typed(boolean, default_value=False),
     host="localhost",
     port=typed(int, default_value=11211),
+    connect_timeout=3,
```
GabeLambda: Seconds?

PR #180, opened 2019-09-04
crude implementation

clatko-globe:
simple POC that has much room for optimization.

### DIFF
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -2,6 +2,9 @@
 Process batches of messages.
 
 """
+import asyncio
+import concurrent.futures
+
 from logging import Logger
 from time import time
 from typing import List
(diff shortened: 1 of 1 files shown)


microcosm_pubsub/dispatcher.py
```diff
@@ -36,19 +39,28 @@ def __init__(self, graph):
...
         self.send_batch_metrics = graph.pubsub_send_batch_metrics
         self.max_processing_attempts = graph.config.sqs_message_dispatcher.message_max_processing_attempts
+        self.loop = asyncio.get_event_loop()
 
     def handle_batch(self, bound_handlers) -> List[MessageHandlingResult]:
         """
         Send a batch of messages to a function.
 
         """
+        executor = concurrent.futures.ThreadPoolExecutor(5)
```
clatko-globe: we should add this to __init__ so we are not recreating this over and over. also the magic number should come out.

microcosm_pubsub/dispatcher.py
```diff
@@ -36,19 +39,28 @@ def __init__(self, graph):
...
-        instances = [
-            self.handle_message(message, bound_handlers)
+        instances = []
+
+        threads = [
+            self.loop.run_in_executor(executor, self.handle_message, message, bound_handlers)
             for message in self.sqs_consumer.consume()
         ]
 
+        finished, _ = self.loop.run_until_complete(asyncio.wait(threads, return_when=asyncio.ALL_COMPLETED))
```
clatko-globe: the most inefficient way to wait for everything to complete.

clatko-globe:
i went for the least intrusive way to add concurrency to the daemon, there are many other ways of doing this and many of them involve colored functions - https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/

microcosm_pubsub/dispatcher.py
GabeLambda: Mind expounding what this means a bit?

clatko-globe: i'm using asyncio.ALL_COMPLETED so the arrays returned only contain finished tasks. Another way of doing this is to run within a while true to check on status of tasks and process the ones that finish as they finish.

PR #181, opened 2019-09-06
Better ValidationError logging

adamhadani:
We're currently seeing Hard-To-Debug errors on Loggly raised due to schema validation issues around pubsub messages. These do not include useful information in the stack trace. Minimally, we'd want to know the actual schema (resource) for which the error was raised. This PR adds such functionality. We can add additional useful meta-data in the future using this mechanism as well.

* Adds the schema name as part of the `ValidationError.messages` payload which gets displayed in Loggly
* Update unit-tests to self-check expected behavior

### DIFF
--- a/microcosm_pubsub/codecs.py
+++ b/microcosm_pubsub/codecs.py
@@ -15,6 +15,25 @@ from marshmallow import (
 DEFAULT_MEDIA_TYPE = "application/json"
 
 
+def enrich(error: ValidationError, schema: Schema):
+    """
+    Enrich a schema validation error with a human-interpretable
+    description of a marshmallow schema.
+
+    """
+    messages = error.messages
+    if isinstance(messages, dict):
+        messages.update(dict(
+            schema=str(schema),
+        ))
+    else:
+        messages.append(
+            str(schema),
+        )
+
+    return error
+
+
 class PubSubMessageSchema(Schema):
     """
     Base schema for messages, including a media type.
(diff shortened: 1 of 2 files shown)


microcosm_pubsub/codecs.py
```diff
@@ -15,6 +15,25 @@
...
+    if isinstance(messages, dict):
+        messages.update(dict(
+            schema=str(schema),
+        ))
+    else:
+        messages.append(
+            str(schema),
+        )
+
+    return error
```
adamhadani: note that:
1. marshmallow ValidationError uses a `messages` list/dict rather than a `message` str
2. we return the same error object we got passed in, and update its error messages in-place. This is so we dont lose additional meta-data that may have been passed to the original error's [constructor](https://marshmallow.readthedocs.io/en/stable/api_reference.html#marshmallow.exceptions.ValidationError).

microcosm_pubsub/tests/test_codecs.py
```diff
@@ -103,4 +103,7 @@ def test_decode_missing_field():
     message = dumps({
         "mediaType": DerivedSchema.MEDIA_TYPE,
     })
-    assert_that(calling(codec.decode).with_args(message), raises(ValidationError))
+    assert_that(
+        calling(codec.decode).with_args(message),
+        raises(ValidationError, r".*DerivedSchema"),
```
adamhadani: slight update to unit-test to confirm part of exception payload now includes schema name

PR #182, opened 2019-09-17
add default ttl

clatko-globe:
do we not want to set cache on cache hit?

### DIFF
--- a/microcosm_pubsub/constants.py
+++ b/microcosm_pubsub/constants.py
@@ -3,6 +3,7 @@ Constants.
 
 """
 
+DEFAULT_RESOURCE_CACHE_TTL = 60
 PUBLISHED_KEY = "X-Request-Published"
 TTL_KEY = "X-Request-Ttl"
 URI_KEY = "uri"
--- a/microcosm_pubsub/handlers/uri_handler.py
+++ b/microcosm_pubsub/handlers/uri_handler.py
@@ -13,6 +13,9 @@ from microcosm_pubsub.conventions.lifecycle import LifecycleChange
 from microcosm_pubsub.errors import Nack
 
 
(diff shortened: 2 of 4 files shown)


microcosm_pubsub/constants.py
```diff
@@ -3,6 +3,7 @@
 
 """
 
+DEFAULT_RESOURCE_CACHE_TTL = 60
```
adamhadani: @clatko-globe FYI moved this to `constants.py` file

microcosm_pubsub/handlers/uri_handler.py
```diff
@@ -60,11 +61,13 @@ def __init__(
         retry_nack_timeout=1,
         resource_nack_timeout=1,
         resource_cache_enabled=True,
+        resource_cache_ttl=DEFAULT_RESOURCE_CACHE_TTL,
```
adamhadani: @clatko-globe FYI changed this to be a simple constructor argument rather than a `graph.resource_cache` config parameter.. thinking is:
a. simpler signature here without having to access some other component's config block
b. in particular this lets us tune resource cache ttl *per handler*, which could be useful.

setup.py
```diff
@@ -21,7 +21,7 @@
         "dataclasses;python_version<'3.7'",
         "marshmallow>=3.0.0",
         "microcosm>=2.12.0",
-        "microcosm-caching>=0.1.0",
+        "microcosm-caching>=0.2.0",
```
adamhadani: @clatko-globe  FYI bumping this to make sure we include support for `ttl` parameter, which I added after the `0.1.0` release

clatko-globe: cool. thanks for the comments.

PR #183, opened 2019-10-09
Rework de/serialization for pubsub messages

GabeLambda:
The current method used for de/serializing messages is a bit obtuse at
the moment, and prevents some amount of ability to use some enhanced
messages (e.g. enums).

Note that this change may not backwards compatible, producing services
may start creating messages that will fail decoding. Edit: From the looks of it, this is in line with what's already being published in the message today (i.e. camelCased fields). This is just a possible warning here.

### DIFF
--- a/microcosm_pubsub/codecs.py
+++ b/microcosm_pubsub/codecs.py
@@ -3,6 +3,7 @@ Message encoding and decoding.
 
 """
 from json import dumps, loads
+from typing import Any, Dict, Optional
 
 from marshmallow import (
     EXCLUDE,
@@ -39,38 +40,16 @@ class PubSubMessageSchema(Schema):
     Base schema for messages, including a media type.
 
     """
-    media_type = fields.Method(
-        # called by dump
-        serialize="serialize_media_type",
-        # called by load
-        deserialize="deserialize_media_type",
-        attribute="mediaType",
-        # need to set missing to non-None or marshmallow won't call the deserialize function
-        missing=DEFAULT_MEDIA_TYPE,
-    )
-    opaque_data = fields.Dict(
-        attribute="opaqueData",
+    MEDIA_TYPE = DEFAULT_MEDIA_TYPE
+
+    mediaType = fields.String(
+        attribute="media_type",
         required=False,
     )
-
-    def serialize_media_type(self, message):
-        """
-        Fetch the media type from the message.
-
-        """
-        try:
-            return message["mediaType"]
-        except KeyError:
-            raise ValidationError("Message did not define a media type")
-
-    def deserialize_media_type(self, obj):
-        """
-        Return a custom media type.
-
-        Should be overridden in subclasses.
-
-        """
-        return DEFAULT_MEDIA_TYPE
+    opaqueData = fields.Dict(
+        attribute="opaque_data",
+        required=False,
+    )
 
 
 class MediaTypeSchema(Schema):
@@ -87,10 +66,10 @@ class PubSubMessageCodec:
 
     """
 
-    def __init__(self, schema):
+    def __init__(self, schema: PubSubMessageSchema):
         self.schema = schema
 
-    def encode(self, dct=None, **kwargs):
+    def encode(self, dct: Optional[Dict[str, Any]] = None, **kwargs) -> str:
         """
         Encode a message.
 
@@ -99,10 +78,16 @@ class PubSubMessageCodec:
         """
         message = dct.copy() if dct else dict()
         message.update(kwargs)
-        # We allow unknown fields to pass through (but not included here)
-        # to accommodate unconditional parameter passing during produce()
-        # e.g. passing in `uri` for IdentityMessages
-        return dumps(self.schema.load(message, unknown=EXCLUDE))
+
+        dumped = self.schema.dump(message)
+
+        # NB: Schema dump doesn't run validation
+        errors = self.schema.validate(dumped)
+        if errors:
+            raise ValidationError(message=errors)
+
+        # Dump to string
+        return dumps(dumped)
 
     def decode(self, message):
         """
@@ -119,9 +104,7 @@ class PubSubMessageCodec:
         try:
             # load performs a validation and raises on error
             # Similarly, we exclude unknown fields to avoid errors on decode
-            self.schema.load(dct, unknown=EXCLUDE)
+            return self.schema.load(dct, unknown=EXCLUDE)
         except ValidationError as error:
             # Add more useful information to logs for debugging validation errors.
             raise enrich(error, self.schema)
-
-        return self.schema.dump(dct)
--- a/microcosm_pubsub/codecs.py
+++ b/microcosm_pubsub/codecs.py
@@ -44,7 +44,7 @@ class PubSubMessageSchema(Schema):
 
     mediaType = fields.String(
         attribute="media_type",
-        required=False,
+        required=True,
     )
     opaqueData = fields.Dict(
         attribute="opaque_data",
--- a/microcosm_pubsub/matchers/message.py
+++ b/microcosm_pubsub/matchers/message.py
@@ -76,7 +76,7 @@ class PublishedMessage:
         Iterate over published messages from a mocked SNSProducer.
 
         """
-        if isinstance(sns_producer, Mock):
+        if isinstance(sns_producer, Mock):  # type: ignore
             target = sns_producer.produce
         else:
             target = sns_producer.sns_client.publish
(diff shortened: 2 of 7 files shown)


microcosm_pubsub/codecs.py
```diff
@@ -39,39 +40,17 @@ class PubSubMessageSchema(Schema):
...
-        # called by dump
-        serialize="serialize_media_type",
-        # called by load
-        deserialize="deserialize_media_type",
-        attribute="mediaType",
-        # need to set missing to non-None or marshmallow won't call the deserialize function
-        missing=DEFAULT_MEDIA_TYPE,
+    MEDIA_TYPE = DEFAULT_MEDIA_TYPE
+
+    mediaType = fields.String(
```
GabeLambda: One piece changed here is to make encoding/decoding symmetrical and in-line with how we already do it in other libraries, like in microcosm-flask. This should make it explicit that:
* messages sent over the wire are camelCased
* loaded messages have snaked_cased attributes

microcosm_pubsub/codecs.py
```diff
@@ -39,39 +40,17 @@ class PubSubMessageSchema(Schema):
...
 
-    def serialize_media_type(self, message):
-        """
-        Fetch the media type from the message.
-
-        """
-        try:
-            return message["mediaType"]
-        except KeyError:
-            raise ValidationError("Message did not define a media type")
```
GabeLambda: There was an odd level indecision here regarding whether or not all published messages required a media_type, and whether it should be present at the time of message encoding. I'm taking the stance here that all messages require the media type to be explicitly-specified prior to creation, and treated as such (i.e no implicit loading of media types).

microcosm_pubsub/codecs.py
```diff
@@ -99,10 +78,16 @@ def encode(self, dct=None, **kwargs):
         """
         message = dct.copy() if dct else dict()
         message.update(kwargs)
-        # We allow unknown fields to pass through (but not included here)
-        # to accommodate unconditional parameter passing during produce()
-        # e.g. passing in `uri` for IdentityMessages
-        return dumps(self.schema.load(message, unknown=EXCLUDE))
```
GabeLambda: This was the main crux of the issue I was describing wrt lack of type freedom:
* we were sending the decoded version of the resource over the wire
* this means that custom types that defined a de/serialization method wouldn't even have the chance to run
* this is brain-melting; the encoded form is what's transferred, the decoded form is what's parsed. We shouldn't be confusing the two.

microcosm_pubsub/tests/conventions/test_messages.py
```diff
@@ -23,13 +25,72 @@
...
+    key = auto()
+
+    def __str__(self):
+        return self.name
+
+
+class EnumField(Field):
+    """
+    Test serialization of non-serializable fields
+    This is mostly taken from microcosm-flask's EnumField
```
GabeLambda: Either types are added to their own lib, or we depend on mc-flask. I opted for copying.

microcosm_pubsub/tests/conventions/test_messages.py
```diff
@@ -102,17 +189,17 @@ def test_decode_identity_message_schema():
     Message decoding should process standard IdentityMessage fields.
 
     """
-    schema = IdentityMessageSchema(make_media_type("Foo"))
-    codec = PubSubMessageCodec(schema)
+    identity_schema = IdentityMessageSchema(make_media_type("Foo"))
+    codec = PubSubMessageCodec(identity_schema)
     message = dumps({
-        "mediaType": "application/vnd.globality.pubsub.foo",
+        "mediaType": "application/vnd.globality.pubsub._.created.foo",
```
GabeLambda: This previously made no sense; if we were decoding a message that was associated with a specific message, we should be validating that media type.

Come to think of it, we should probably do that validation during load.

microcosm_pubsub/matchers/message.py
```diff
@@ -76,7 +76,7 @@ def iter_from_sns_producer(sns_producer) -> Iterable["PublishedMessage"]:
         Iterate over published messages from a mocked SNSProducer.
 
         """
-        if isinstance(sns_producer, Mock):
+        if isinstance(sns_producer, Mock):  # type: ignore
```
GabeLambda: This was raising a new error (`Parameterized generics cannot be used with class or instance checks`) (this is also on develop, btw).

microcosm_pubsub/codecs.py
```diff
@@ -87,10 +66,10 @@ class PubSubMessageCodec:
 
     """
 
-    def __init__(self, schema):
+    def __init__(self, schema: PubSubMessageSchema):
```
afallou: By using typing here we're implicitly saying we're not supporting Python pre-3.6. Is that something we want?

microcosm_pubsub/codecs.py
```diff
@@ -99,10 +78,16 @@ def encode(self, dct=None, **kwargs):
...
         message.update(kwargs)
-        # We allow unknown fields to pass through (but not included here)
-        # to accommodate unconditional parameter passing during produce()
-        # e.g. passing in `uri` for IdentityMessages
-        return dumps(self.schema.load(message, unknown=EXCLUDE))
+
+        dumped = self.schema.dump(message)
+
+        # NB: Schema dump doesn't run validation
+        errors = self.schema.validate(dumped)
```
afallou: This bit is a bit surprising, the fact that we have to run `validate` on `dumped` rather than `message`

microcosm_pubsub/codecs.py
```diff
@@ -119,9 +104,7 @@ def decode(self, message):
         try:
             # load performs a validation and raises on error
             # Similarly, we exclude unknown fields to avoid errors on decode
-            self.schema.load(dct, unknown=EXCLUDE)
+            return self.schema.load(dct, unknown=EXCLUDE)
         except ValidationError as error:
             # Add more useful information to logs for debugging validation errors.
             raise enrich(error, self.schema)
-
-        return self.schema.dump(dct)
```
afallou: Yeah that was weird

microcosm_pubsub/codecs.py
```diff
@@ -39,39 +40,17 @@ class PubSubMessageSchema(Schema):
...
     """
-    media_type = fields.Method(
-        # called by dump
-        serialize="serialize_media_type",
-        # called by load
-        deserialize="deserialize_media_type",
-        attribute="mediaType",
-        # need to set missing to non-None or marshmallow won't call the deserialize function
-        missing=DEFAULT_MEDIA_TYPE,
+    MEDIA_TYPE = DEFAULT_MEDIA_TYPE
```
afallou: Is `MEDIA_TYPE` an attribute of the base class that we're overriding here?

microcosm_pubsub/tests/conventions/test_messages.py
afallou: Do you mean "a message that was associated with a specific media type"?

GabeLambda: True, I'm probably misunderstanding some usage here. When a message is deserialized, we use this key to determine which schema to decode it with; it doesn't make sense, then, that the test is using a mismatched media type.
So yeah, validation not needed; we get that via the registry.

microcosm_pubsub/codecs.py
GabeLambda: That was already done when we started using mm-3.0.

microcosm_pubsub/codecs.py
GabeLambda: It should be overridden on all subclasses, yeah. Either that, or it can be made an abstract property of the base class, and overridden for all subclasses.

microcosm_pubsub/codecs.py
GabeLambda: They're both dicts. The only difference is that the `message` is the `object` form, and `dumped` is the "pre-serialized" form (e.g. all enums cast to str, all fields converted to the wire form).

Running validate on `message` will error out, because the fields haven't yet been converted by the dump.

`validate` essentially ensures that you didn't forget to pass in data that needed to be there (e.g. didn't supply a required field).

microcosm_pubsub/codecs.py
```diff
@@ -39,39 +40,17 @@ class PubSubMessageSchema(Schema):
...
+
+    mediaType = fields.String(
+        attribute="media_type",
+        required=False,
```
cdm-ium: So can it be `required=true`?

microcosm_pubsub/codecs.py
cdm-ium: Gabe is right, this ship has sailed. We're never going back.

microcosm_pubsub/codecs.py
GabeLambda: You know, probably.

GabeLambda: I'll leave it as-is for the moment.

PR #185, opened 2019-10-16
Use dictionary syntax instead of constructor

afallou:
The constructor syntax `dict(**a, **b)` will error if `a` and `b` share keys, where the `{}` syntax will simply take the value from the second item.

### DIFF
--- a/microcosm_pubsub/result.py
+++ b/microcosm_pubsub/result.py
@@ -146,7 +146,7 @@ class MessageHandlingResult:
             self.result.level,
             entry,
             exc_info=self.exc_info,
-            extra=dict(**self.extra, **opaque.as_dict()),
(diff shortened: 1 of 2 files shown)


microcosm_pubsub/result.py
```diff
@@ -146,7 +146,7 @@ def log(self, logger, opaque):
             self.result.level,
             entry,
             exc_info=self.exc_info,
-            extra=dict(**self.extra, **opaque.as_dict()),
+            extra={**opaque.as_dict(), **self.extra},
```
afallou: I'm switching the order so that we end up with the values that were explicitly set (from `extra`) rather than the ones implicitly defined (from `opaque`).

GabeLambda:
Test?

PR #187, opened 2019-10-17
Release version 2.3.1

afallou:
This will fix the errors we're seeing with `object got two values for keyword argument: uri`.

### DIFF
--- a/microcosm_pubsub/matchers/message.py
+++ b/microcosm_pubsub/matchers/message.py
@@ -76,7 +76,7 @@ class PublishedMessage:
         Iterate over published messages from a mocked SNSProducer.
 
         """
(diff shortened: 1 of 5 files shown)


microcosm_pubsub/matchers/message.py
```diff
@@ -76,7 +76,7 @@ def iter_from_sns_producer(sns_producer) -> Iterable["PublishedMessage"]:
         Iterate over published messages from a mocked SNSProducer.
 
         """
-        if isinstance(sns_producer, Mock):
+        if isinstance(sns_producer, Mock):  # type: ignore
```
afallou: Apparently mypy has gotten stricter. Ignoring this since it's a test matcher.

PR #189, opened 2019-11-04
Remove media type for convention-driven message schema

afallou:
**Why?**
A media type was being passed in to our convention-driven message schemas. This was a hacky way of making them work with the pubsub registry; this has since been fixed by allowing the registry to store media types without an associated schema, in which case it defaults back to convention.

Having a required constructor argument for a schema is bad practice in my opinion, especially since that argument is not actually needed to define the schema.

**What?**
Remove `media_type` from constructor for `URIMessageSchema` and `IdentityMessageSchema`.

### DIFF
--- a/microcosm_pubsub/conventions/messages.py
+++ b/microcosm_pubsub/conventions/messages.py
@@ -14,10 +14,6 @@ class URIMessageSchema(PubSubMessageSchema):
     the consumer can retry a few times before giving up (e.g. via SQS dead-lettering).
 
     """
-    def __init__(self, media_type, **kwargs):
-        super().__init__(**kwargs)
-        self.MEDIA_TYPE = media_type
-
     uri = fields.String(required=True)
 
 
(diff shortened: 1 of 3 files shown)


microcosm_pubsub/conventions/messages.py
```diff
@@ -14,10 +14,6 @@ class URIMessageSchema(PubSubMessageSchema):
     the consumer can retry a few times before giving up (e.g. via SQS dead-lettering).
 
     """
-    def __init__(self, media_type, **kwargs):
-        super().__init__(**kwargs)
-        self.MEDIA_TYPE = media_type
```
afallou: There is an argument for making `media_type` an optional argument for backward compatibility. However, I believe this hasn't been needed since early 2017; I'd rather clean this up properly.

microcosm_pubsub/registry.py
```diff
@@ -77,11 +77,11 @@ def find(self, media_type):
...
             # use convention otherwise
             if self.lifecycle_change.Deleted in media_type.split("."):
-                schema = IdentityMessageSchema(media_type)
+                schema = IdentityMessageSchema()
             elif self.lifecycle_change.Changed in media_type.split("."):
-                schema = ChangedURIMessageSchema(media_type)
+                schema = ChangedURIMessageSchema()
             else:
-                schema = URIMessageSchema(media_type)
+                schema = URIMessageSchema()
```
afallou: By this point we have found the schema we're going to be using for the message, based on the media type. This means that we won't need the media type again.

PR #197, opened 2019-11-06
Wrap producer in metrics

KensoDev:
[https://globality.atlassian.net/browse/DEVOPS-636](https://globality.atlassian.net/browse/DEVOPS-636)

### DIFF
--- a/microcosm_pubsub/metrics.py
+++ b/microcosm_pubsub/metrics.py
@@ -120,3 +120,52 @@ class PubSubSendBatchMetrics:
             message_batch_size,
             tags=tags,
         )
+
+
+@defaults(
+    enabled=typed(boolean, default_value=True)
+)
+class PubSubProducerMetrics:
+    """
+    Send metrics regarding the producer
+
+    """
+
+    def __init__(self, graph):
+        self.metrics = self.get_metrics(graph)
+        self.enabled = bool(
+            self.metrics
+            and self.metrics.host != "localhost"
+            and graph.config.pubsub_send_metrics.enabled
+        )
+
+    def get_metrics(self, graph):
+        """
+        Fetch the metrics client from the graph.
+
+        Metrics will be disabled if the not configured.
+
+        """
+        try:
+            return graph.metrics
+        except NotBoundError:
+            return None
+
+    def __call__(self, elapsed_time: float, publish_result: bool, **kwargs):
+        """
+        Send metrics for how long it takes to produce a message
+
+        """
+        if not self.enabled:
+            return
+
+        tags = [
+            "source:microcosm-pubsub",
+            f"publish_result:{publish_result}",
+        ]
+
+        self.metrics.histogram(
+            "sns_producer",
+            elapsed_time,
+            tags=tags,
+        )
--- a/microcosm_pubsub/producer.py
+++ b/microcosm_pubsub/producer.py
@@ -130,7 +130,7 @@ class SNSProducer:
         )
 
         if publish_exception:
-            raise Exception(f"Could not publish message, SNS producer error")
+            raise Exception(f"Could not publish message, SNS producer error: {publish_exception}")
 
         self.logger.debug("Published message with media type {media_type}", extra=extra)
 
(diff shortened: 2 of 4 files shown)


microcosm_pubsub/producer.py
```diff
@@ -100,16 +110,32 @@ def publish_message(self, pubsub_message: PubsubMessage):
...
-                Message=pubsub_message.message,
-                MessageAttributes=pubsub_message.message_attributes,
-            )
+            try:
+                result = self.sns_client.publish(
+                    TopicArn=pubsub_message.topic_arn,
+                    Message=pubsub_message.message,
+                    MessageAttributes=pubsub_message.message_attributes,
+                )
+            except Exception as e:
```
KensoDev: Tracking the botocore exception here (after the timeout is done).
We catch it here but throw it later after we send the stats

microcosm_pubsub/producer.py
```diff
@@ -100,16 +110,32 @@ def publish_message(self, pubsub_message: PubsubMessage):
             **pubsub_message.opaque_data
         )
         self.logger.debug("Publishing message with media type {media_type}", extra=extra)
+        publish_result = "SUCCESS"
```
KensoDev: This is sort of iffy, I know.

microcosm_pubsub/producer.py
```diff
@@ -100,16 +110,32 @@ def publish_message(self, pubsub_message: PubsubMessage):
...
+                publish_exception = e
+                publish_result = "FAILURE"
+
+        self.pubsub_producer_metrics(
+            publish_result=publish_result,
+            **extra
+        )
+
+        if publish_exception:
+            raise Exception(f"Could not publish message, SNS producer error")
```
GabeLambda: Why not just re-raise the error?

microcosm_pubsub/metrics.py
```diff
@@ -120,3 +120,52 @@ def __call__(self, elapsed_time: float, message_batch_size: int):
...
+    Send metrics regarding the producer
+
+    """
+
+    def __init__(self, graph):
+        self.metrics = self.get_metrics(graph)
+        self.enabled = bool(
+            self.metrics
+            and self.metrics.host != "localhost"
+            and graph.config.pubsub_send_metrics.enabled
```
pankajghosh: Should this be pubsub_producer_metrics.enabled?

KensoDev: No. This is the generic one that we send in ansible.
I don't want to have a fine grained one. If we track pubsub metrics, it should track dispatcher and producer.

microcosm_pubsub/producer.py
KensoDev: No reason outside of logging search

microcosm_pubsub/metrics.py
pankajghosh: Sounds good.

microcosm_pubsub/producer.py
GabeLambda: Then why not:
```
             raise Exception(f"Could not publish message, SNS producer error: {publish_exception}")
```
I for one would like to know what the actual problem was.

microcosm_pubsub/metrics.py
```diff
@@ -120,3 +120,52 @@ def __call__(self, elapsed_time: float, message_batch_size: int):
...
+            return
+
+        tags = [
+            "source:microcosm-pubsub",
+            f"publish_result:{publish_result}",
+        ]
+
+        self.metrics.histogram(
+            "sns_producer",
+            elapsed_time,
```
GabeLambda: Nit: convert to ms?

microcosm_pubsub/producer.py
KensoDev: Done!

GabeLambda:
Is there a reason this couldn't be rolled into the original publish metrics?

I'm wondering if we couldn't record the metrics here:
https://github.com/globality-corp/microcosm-pubsub/blob/89b118590d052433a2f3175a64e03c47e26121b2/microcosm_pubsub/result.py#L71

https://github.com/globality-corp/microcosm-pubsub/blob/develop/microcosm_pubsub/metrics.py#L37

That probably does end up adding a bit more work, at the benefit of having less exception handling.

PR #200, opened 2019-12-18
Add diagram of handling flow

afallou:
This is to make it easier to understand the various handling outcomes and how they interact with SQS.

### DIFF
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -1,6 +1,31 @@
 """
 Process batches of messages.
 
+Flow diagram:
+
++-----------------------+                        +-------------+                                            +----------+
+|                       |                        |             |                                            |          |
+|                       |     1. Get messages    |             |                                            |          |
+| SQSMessageDispatcher  |<-----------------------+             +<-------------------------------------------+          |
+|                       |                        |             |                                            |          |
+|                       |                        |             |                                            |          |
++-------+-+-------------+                        |             |                                            |          |
+           | ˆ                                   |             |                                            |          |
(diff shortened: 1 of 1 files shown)


microcosm_pubsub/dispatcher.py
```diff
@@ -1,6 +1,31 @@
 """
 Process batches of messages.
 
+Flow diagram:
+
++-----------------------+                        +-------------+                                            +----------+
+|                       |                        |             |                                            |          |
+|                       |     1. Get messages    |             |                                            |          |
```
german-glob: So the `SQSMessageDispatcher` gets messages from `SQSConsumer` or vice-versa? I'm asking because arrows point to both nodes.

GabeLambda:
This might be clearer with an actual image vs. ascii art.

microcosm_pubsub/dispatcher.py
GabeLambda: The dispatcher gets messages from the SQS consumer.

I'm guessing "propagate result" means `ack`/`nack`? Then that's the second part of that bidirectional flow.

I think making that arrow go both ways is a bit confusing however. I think if you removed that (indicating that the dispatcher takes messages from the consumer, handles/acks them), then "get messages" could be assumed to be the next part of the dispatchers lifecycle.

afallou: Updated arrow direction.
As for "propagate result", I moved it to have the explanation directly underneath.

afallou:
@GabeLambda sure, but I can't put an image in the code. And in any case ascii art beats no documentation 😉

PR #201, opened 2019-12-19
Improve error message handling during chain handler execution

GabeLambda:
Reading chain errors tends to be a bit cryptic, mostly due to the fact
that the actual calling function has been funnelled/wrapped so many
times that it may be unclear exactly where an error has arisen.

Example:
```
  File "microcosm_pubsub/handlers/chain_handlers.py", line 42, in handle
    chain(**kwargs)
  File "microcosm_pubsub/chain/chain.py", line 68, in __call__
    res = func()
  File "microcosm_pubsub/chain/context_decorators.py", line 38, in decorate
    for arg_name, default in positional_args[len(args):]
  File "microcosm_pubsub/chain/context_decorators.py", line 39, in <dictcomp>
    if arg_name not in kwargs
  File "microcosm_pubsub/chain/context.py", line 20, in __getitem__
    return self.store[key]
KeyError: 'context_key'
```

Additionally, failures to find an attribute during `assign` calls end up
also missing helpful context:
```
  File "microcosm-pubsub/microcosm_pubsub/chain/statements/assign.py", line 28, in __call__
    value = value[part]
  File "site-packages/bravado_core/model.py", line 435, in __getitem__
    return self.__dict[property_name]
KeyError: 'id'
```

This changeset updates errors such that:
* On the inability to find a context key when evaluating a chained
function, both the key and the function are reported in the error
Example:
```
microcosm_pubsub.chain.exceptions.ContextKeyNotFound: Failed to find context_key `key` during evaluation of `func`
```

* If an attribute is unable to be found during assignment, both the
field and the parent are dutifully reported.
Example:
```
microcosm_pubsub.chain.exceptions.AttributeNotFound: Failed to find attribute `id` on `event`
```

### DIFF
--- a/microcosm_pubsub/chain/chain.py
+++ b/microcosm_pubsub/chain/chain.py
@@ -63,6 +63,8 @@ class Chain:
         context = context or self.new_context_type()
         context.update(kwargs)
 
+        res = None
+
         for link in self.links:
             func = self.apply_decorators(context, link)
             res = func()
--- a/microcosm_pubsub/chain/context_decorators.py
+++ b/microcosm_pubsub/chain/context_decorators.py
@@ -2,6 +2,7 @@ from functools import WRAPPER_ASSIGNMENTS, wraps
 from inspect import Parameter, Signature, signature
 
 from microcosm_pubsub.chain.decorators import BINDS, EXTRACTS
+from microcosm_pubsub.chain.exceptions import ContextKeyNotFound
 
 
 DEFAULT_ASSIGNED = (EXTRACTS, BINDS)
@@ -33,11 +34,15 @@ def get_from_context(context, func, assigned=DEFAULT_ASSIGNED):
 
     @wraps(func, assigned=assigned + WRAPPER_ASSIGNMENTS)
     def decorate(*args, **kwargs):
-        context_kwargs = {
-            arg_name: (context[arg_name] if default is Signature.empty else context.get(arg_name, default))
-            for arg_name, default in positional_args[len(args):]
-            if arg_name not in kwargs
-        }
+        try:
+            context_kwargs = {
+                arg_name: (context[arg_name] if default is Signature.empty else context.get(arg_name, default))
+                for arg_name, default in positional_args[len(args):]
+                if arg_name not in kwargs
+            }
+        except KeyError as error:
+            raise ContextKeyNotFound(error, func)
+
         return func(*args, **kwargs, **context_kwargs)
     return decorate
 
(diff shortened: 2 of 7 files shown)


microcosm_pubsub/chain/chain.py
```diff
@@ -63,6 +63,8 @@ def __call__(self, context=None, **kwargs):
         context = context or self.new_context_type()
         context.update(kwargs)
 
+        res = None
```
GabeLambda: Technically, if there are no `self.links`, this var could be undefined when we try and return out. Default to null instead.

microcosm_pubsub/chain/exceptions.py
```diff
@@ -0,0 +1,10 @@
+class AttributeNotFound(Exception):
+    def __init__(self, parent, attribute):
+        super().__init__(f"Failed to find attribute `{attribute}` on `{parent}`")
+
+
+class ContextKeyNotFound(Exception):
+    def __init__(self, key_error: KeyError, func):
+        context_key, *_ = key_error.args
```
afallou: I had no idea `*` could be used on the left-hand side like this.

microcosm_pubsub/chain/exceptions.py
```diff
@@ -0,0 +1,10 @@
+class AttributeNotFound(Exception):
+    def __init__(self, parent, attribute):
+        super().__init__(f"Failed to find attribute `{attribute}` on `{parent}`")
```
afallou: Worth adding a test for those error messages?

microcosm_pubsub/chain/context_decorators.py
```diff
@@ -33,11 +34,15 @@ def get_from_context(context, func, assigned=DEFAULT_ASSIGNED):
...
-            if arg_name not in kwargs
-        }
+        try:
+            context_kwargs = {
+                arg_name: (context[arg_name] if default is Signature.empty else context.get(arg_name, default))
+                for arg_name, default in positional_args[len(args):]
+                if arg_name not in kwargs
+            }
+        except KeyError as error:
+            raise ContextKeyNotFound(error, func)
```
mlcohen: Fantastic 👍🏼

microcosm_pubsub/chain/exceptions.py
mlcohen: 😮

microcosm_pubsub/chain/exceptions.py
GabeLambda: Hrm, sure.

microcosm_pubsub/chain/exceptions.py
GabeLambda: Yeah. It's a python3-ism.

PR #203, opened 2020-01-02
Configuration to avoid fetching resource

afallou:
**Why?**
This is part of POC to transition to handlers that don't fetch the published resource, and instead only use the message itself.

**What?**
- Add configuration value to bypass fetching the resource in URIHandler
- Add base handler that has the same behavior, and that will be used instead of URIHandler in the future.

### DIFF
--- a/microcosm_pubsub/handlers/base.py
+++ b/microcosm_pubsub/handlers/base.py
@@ -1,0 +1,160 @@
+"""
+Base pubsub handler
+
+"""
+from abc import ABCMeta
+from re import search
+
+from inflection import titleize
+from microcosm.errors import LockedGraphError, NotBoundError
+
+from microcosm_pubsub.constants import DEFAULT_RESOURCE_CACHE_TTL
+from microcosm_pubsub.conventions.lifecycle import LifecycleChange
+
+
+def resource_cache_whitelist_callable(media_type, uri):
+    """
+    Default resource cache whitelist callable implementation.
+
+    Only whitelists under the following conditions:
+    * verb is created
+    * resource is an event
+
+    """
+    if not media_type:
+        # Nb. likely to happen with mocks and tests.
+        return False
+
+    return all((
+        search(r"/[a-z]+_event/", uri),
+        search(r".{}.".format(LifecycleChange.Created), media_type),
+    ))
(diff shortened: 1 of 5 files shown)


microcosm_pubsub/handlers/base.py
```diff
@@ -0,0 +1,160 @@
...
+        # Nb. likely to happen with mocks and tests.
+        return False
+
+    return all((
+        search(r"/[a-z]+_event/", uri),
+        search(r".{}.".format(LifecycleChange.Created), media_type),
+    ))
+
+
+class PubSubHandler(metaclass=ABCMeta):
```
afallou: Everything in this class was taken from `URIHandler`.

microcosm_pubsub/handlers/uri_handler.py
```diff
@@ -42,121 +17,33 @@ class URIHandler(metaclass=ABCMeta):
...
 
-        resource = self.convert_resource(
-            self.get_resource(message, uri),
-        )
+        resource = None
+        # XXX: remove conditional and use base class instead post-POC
+        if self.fetch_resource:
+            resource = self.convert_resource(
+                self.get_resource(message, uri),
+            )
```
afallou: Those lines are the only difference with the base class for the `__call__` method.

microcosm_pubsub/handlers/base.py
```diff
@@ -0,0 +1,160 @@
...
+        resource_cache_whitelist_callable=resource_cache_whitelist_callable,
+    ):
+        self.opaque = graph.opaque
+        self.retry_nack_timeout = retry_nack_timeout
+        self.resource_nack_timeout = resource_nack_timeout
+        self.resource_cache_ttl = resource_cache_ttl
+        self.resource_cache = self.get_resource_cache(graph) if resource_cache_enabled else None
+        self.resource_cache_whitelist_callable = resource_cache_whitelist_callable
+
+    @property
```
GabeLambda: We can probably just remove this if we're going all-new here.

afallou: I'll move it to `URIHandler` instead of the base class.

PR #204, opened 2020-01-02
Release 2.9.0

afallou:
Releases https://github.com/globality-corp/microcosm-pubsub/pull/203
**Why?**
This is part of POC to transition to handlers that don't fetch the published resource, and instead only use the message itself.

**What?**
- Add configuration value to bypass fetching the resource in URIHandler
- Add base handler that has the same behavior, and that will be used instead of URIHandler in the future.

### DIFF
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -19,7 +19,8 @@ from microcosm_pubsub.result import MessageHandlingResult, MessageHandlingResult
 @logger
 @defaults(
     # Number of failed attempts after which the message stops being processed
-    message_max_processing_attempts=typed(int, default_value=None)
+    message_max_processing_attempts=typed(int, default_value=None),
+    fetch_uri_resource=typed(bool, default_value=True),
 )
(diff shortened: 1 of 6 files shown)


microcosm_pubsub/dispatcher.py
```diff
@@ -19,7 +19,8 @@
 @logger
 @defaults(
     # Number of failed attempts after which the message stops being processed
-    message_max_processing_attempts=typed(int, default_value=None)
+    message_max_processing_attempts=typed(int, default_value=None),
```
sethisernhagen: What does this do? Does this behave differently than SQS's `maxReceiveCount` config?

afallou: That's an old thing that was put in for local/integration testing. In those cases we use LocalStack, which doesn't deadletter. Instead we keep a count of the number of times the message was processed. It's unused by default though.

afallou:
@mlcohen the full description is in the linked PR; I copy-pasted it here (which doesn't hurt, I admit).

PR #205, opened 2020-01-03
Make fetch-less handling backwards compatible

afallou:
Avoid changing method signature.


microcosm_pubsub/handlers/base.py
```diff
@@ -151,5 +153,5 @@ def get_headers(self, message):
         """
         return self.opaque.as_dict()
 
-    def handle(self, message, uri, **kwargs):
+    def handle(self, message, uri, resource):
```
afallou: I brought the signature of this method and the two above back to what it used to be before https://github.com/globality-corp/microcosm-pubsub/pull/204; for backwards-compatibility.

PR #209, opened 2020-01-06
adding media type to metrics

clatko-globe:
pr straight to master


clatko-globe:
https://github.com/globality-corp/microcosm-pubsub/blob/develop/microcosm_pubsub/producer.py#L109

PR #217, opened 2020-01-11
Add info that will be useful for tacking call patterns.

cdm-ium:
Why? We need better visibility into the call flow between services and
this will help provide that visibility until we have a better tool, e.g.
jaeger.

THIS IS NOT READY FOR MERGE, I am opening to collaborate with @clatko-globe and share some tribal knowledge.

### DIFF
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -28,6 +28,7 @@ Flow diagram:
 
 """
 from logging import Logger
+from os import environ
 from time import time
 from typing import List
 
@@ -110,6 +111,9 @@ class SQSMessageDispatcher:
                 try:
                     self.validate_message(message)
                     handler = self.find_handler(message, bound_handlers)
+                    self.opaque["X-Request-Handler"] = handler.__name__
+                    # XXX tech debt story to fix setting of metrics name
+                    self.opaque["X-Request-Daemon"] = environ.get("METRICS_NAME", "Unknown")
                     instance = MessageHandlingResult.invoke(
                         handler=self.wrap_handler(handler),
                         message=message,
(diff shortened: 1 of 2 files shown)


microcosm_pubsub/dispatcher.py
```diff
@@ -110,6 +111,9 @@ def handle_message(self, message, bound_handlers) -> MessageHandlingResult:
                 try:
                     self.validate_message(message)
                     handler = self.find_handler(message, bound_handlers)
+                    self.opaque["X-Request-Handler"] = handler.__name__
```
cdm-ium: Assigning to opaque directly is probably ok.

We want this info to get logged by web services, so it needs to be passed as a header from marquez in a form that won't get scrubbed by microcosm-flask. That is why we use the "X-Request-" prefix

See:
- https://github.com/globality-corp/marquez/blob/master/marquez/http_client.py#L69
- https://github.com/globality-corp/microcosm-flask/blob/develop/microcosm_flask/context.py#L17

microcosm_pubsub/dispatcher.py
```diff
@@ -110,6 +111,9 @@ def handle_message(self, message, bound_handlers) -> MessageHandlingResult:
...
+                    # XXX tech debt story to fix setting of metrics name
+                    self.opaque["X-Request-Daemon"] = environ.get("METRICS_NAME", "Unknown")
```
cdm-ium: This isn't good, but it's a pattern we're grandfathered into from `microcosm-metrics`, here: https://github.com/globality-corp/microcosm-metrics/blob/develop/microcosm_metrics/factories.py#L27

We should go back and fix microcosm metrics to load this information through standard loaders and bind it into the graph to a universally accessibly place so this won't break in environments where we don't use config through environment variables.

PR #219, opened 2020-01-21
Wildcard match for media types

afallou:
**Why?**

A single resource schema can be broadcast using different media
types. For example different lifecycle changes could be broadcast to
different consumers, even though the schema is the same, e.g.
`application/vnd.foo.person.created` and
`application/vnd.foo.person.updated`. In those cases we only want to
register a single schema, and have it match for more than one media type.

**What?**

Allow registering a wildcard schema that matches on several
"concrete" schemas.
- Update registry `find` method.
- Add tests.
- Merge tags from `master` back to `develop`.


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.10.0
+current_version = 2.11.1
```
afallou: I also took this opportunity to merge `master` back to `develop`.

cdm-ium: Thank you.

microcosm_pubsub/registry.py
```diff
@@ -72,7 +93,7 @@ def find(self, media_type):
 
         try:
             # use a concrete schema class if any
-            schema_cls = self._mappings[media_type]
+            schema_cls = self._mappings[matching_media_type]
```
cdm-ium: Will this work properly with prefixes? I think you are potentialy returning a prefix on line 74?

afallou: The assumption here is that the media type that's registered in `self._mappings` the library is either the full media type or the wildcard media type (`application/vnd.foo.*`). So when I pass in `applocation/vnd.foo.bar` I should first "translate" it to the registered wildcard, and only the wildcard from there on.

microcosm_pubsub/tests/test_registry.py
```diff
@@ -86,6 +92,18 @@ def test_find(self):
...
+    def test_find_wildcard(self):
+        schema = self.registry.find("application/vnd.microcosm.wildcard.foo_12")
+        assert_that(schema, is_(instance_of(PubSubMessageCodec)))
+        assert_that(schema.schema, is_(instance_of(WildcardSchema)))
+
+    def test_wildcard_does_not_match_extended(self):
+        # The postfix `*` wildcard doesn't match on `.` separators
+        assert_that(
+            calling(self.registry.find).with_args("application/vnd.microcosm.wildcard.foo.bar"),
+            raises(KeyError),
```
rh-glob: Is this the `raise KeyError("Unregistered media type: {}".format(media_type))` case that is raised?

afallou: Yeah that's the one.

PR #220, opened 2020-01-21
Release 2.12.0

afallou:
Release https://github.com/globality-corp/microcosm-pubsub/pull/219 - see there for more context.


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.10.0
+current_version = 2.12.0
```
afallou: The latest published version is actually `2.11.1`, but we had a kerfuffle with git history.

PR #221, opened 2020-01-21
Exclude unknown field in message schemas

afallou:
Unknown fields don't matter much in the context of pubsub messages; we
should ignore them and not error. We used to have this but implicitly lost
it for deserialization in
https://github.com/globality-corp/microcosm-pubsub/pull/183/files#diff-aeca0c2c58f2b77be52bdd40f71d085eL105

### DIFF
--- a/microcosm_pubsub/registry.py
+++ b/microcosm_pubsub/registry.py
@@ -5,6 +5,7 @@ Registry of SQS message handlers.
 from collections import defaultdict
 from inspect import isclass
 
+from marshmallow import EXCLUDE
 from microcosm.api import defaults
 from microcosm_logging.decorators import logger
 
@@ -94,7 +95,7 @@ class PubSubMessageSchemaRegistry:
         try:
             # use a concrete schema class if any
             schema_cls = self._mappings[matching_media_type]
-            schema = schema_cls()
+            schema = schema_cls(unknown=EXCLUDE)
         except KeyError:
             # use convention otherwise
             if self.lifecycle_change.Deleted in media_type.split("."):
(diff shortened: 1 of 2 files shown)


.bumpversion.cfg
```diff
@@ -1,9 +1,8 @@
 [bumpversion]
-current_version = 2.11.1
+current_version = 2.12.0
```
afallou: I'm merging `master` back to `develop` at the same time.

microcosm_pubsub/registry.py
```diff
@@ -94,7 +95,7 @@ def find(self, media_type):
         try:
             # use a concrete schema class if any
             schema_cls = self._mappings[matching_media_type]
-            schema = schema_cls()
+            schema = schema_cls(unknown=EXCLUDE)
```
afallou: The `unknown` instance variable is used as default when no value is passed in directly, which is typically the case in the serialization [`validate`](https://github.com/globality-corp/microcosm-pubsub/blob/develop/microcosm_pubsub/codecs.py#L85) call (for which `unknown` cannot be passed in as an argument).

GabeLambda:
Unit test?

afallou:
@GabeLambda added

microcosm_pubsub/tests/test_registry.py
```diff
@@ -109,6 +109,15 @@ def test_find_changed(self):
         assert_that(schema, is_(instance_of(PubSubMessageCodec)))
         assert_that(schema.schema, is_(instance_of(ChangedURIMessageSchema)))
 
+    def test_serialize_unknown_field(self):
+        schema = self.registry.find(ChangedSchema.MEDIA_TYPE)
+        # No exception raised
+        schema.encode(dict(
```
GabeLambda: Don't need to check that media_type/uri are returned correctly here?

afallou: Not really because we test those things in `test_codecs` already. I'm just testing the added functionality.

PR #225, opened 2020-01-23
Decouple pubsub message production and publishing

afallou:
**Why?**
We have coupled building the content of a message with publishing it, when there is no need to.
The metadata accompanying the message may vary depending on the needs of the publishing library, but the core content of a message won't.

**What?**
- Create message builder component external to `SNSProducer`.
- Mark SNS-specific attribute `topic_arns` and `message_attributes` of `PubsubMessage` as deprecated.
- Change `SNSProducer` code to work with message content being passed in, and keep compatibility with existing mechanism.

### DIFF
--- a/microcosm_pubsub/handlers/publish_batch_message.py
+++ b/microcosm_pubsub/handlers/publish_batch_message.py
@@ -8,7 +8,8 @@ from microcosm_logging.decorators import logger
 
 from microcosm_pubsub.conventions import created
 from microcosm_pubsub.decorators import handles
-from microcosm_pubsub.producer import PubsubMessage, SNSProducer
+from microcosm_pubsub.producer import SNSProducer
+from microcosm_pubsub.published_message import PublishedMessage
 
 
 @binding("publish_message_batch")
@@ -22,7 +23,7 @@ class PublishBatchMessage:
     def __call__(self, message):
         messages = message["messages"]
         for message in messages:
-            pubsub_message = PubsubMessage(
+            pubsub_message = PublishedMessage(
                 media_type=message["media_type"],
                 message=message["message"],
                 message_attributes=message["message_attributes"],
--- a/microcosm_pubsub/producer.py
+++ b/microcosm_pubsub/producer.py
@@ -3,7 +3,6 @@ Message producer.
 
 """
 from collections import defaultdict
-from dataclasses import dataclass
 from distutils.util import strtobool
 from functools import wraps
 from logging import Logger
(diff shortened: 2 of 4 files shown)


microcosm_pubsub/handlers/publish_batch_message.py
```diff
@@ -22,7 +23,7 @@ def __init__(self, graph):
     def __call__(self, message):
         messages = message["messages"]
         for message in messages:
-            pubsub_message = PubsubMessage(
```
afallou: I renamed `PubsubMessage` to `PublishedMessage` because it's an abstraction that's only used on the publisher side.

microcosm_pubsub/producer.py
```diff
@@ -85,40 +78,37 @@ def create_message(self, media_type, dct, uri=None, opaque_data=None, **kwargs)
...
+
         with elapsed_time(extra):
             try:
                 result = self.sns_client.publish(
-                    TopicArn=pubsub_message.topic_arn,
-                    Message=pubsub_message.message,
-                    MessageAttributes=pubsub_message.message_attributes,
+                    TopicArn=topic_arn,
+                    Message=message.message,
+                    MessageAttributes=self.choose_message_attributes(message.media_type),
```
afallou: We can compute `TopicArn` and `MessageAttributes` here, instead of putting them on the pubsub message. Both are SNS-specific.

microcosm_pubsub/published_message.py
```diff
@@ -0,0 +1,55 @@
...
+    topic_arn: str = None
+
+
+@binding("published_message_builder")
+class PublishedMessageBuilder:
+    def __init__(self, graph):
+        self.opaque = graph.opaque
+        self.pubsub_message_schema_registry = graph.pubsub_message_schema_registry
+
+    def __call__(self, media_type, uri=None, opaque_data=None, **kwargs) -> PublishedMessage:
```
afallou: This code is basically a copy of `sns_producer.create_message`, after the pieces about `topic_arns` and `message_attributes` have been removed.

microcosm_pubsub/producer.py
```diff
@@ -201,16 +192,16 @@ def generate_message_batches(self, messages, deferred_batch_size):
...
-                media_type=pubsub_message.media_type,
-                message=pubsub_message.message,
-                message_attributes=pubsub_message.message_attributes,
-                opaque_data=pubsub_message.opaque_data,
-                topic_arn=pubsub_message.topic_arn,
+                media_type=message.media_type,
+                message=message.message,
+                message_attributes=self.producer.choose_message_attributes(message.media_type),
+                opaque_data=message.opaque_data,
+                topic_arn=self.producer.choose_topic_arn(message.media_type),
```
afallou: This implicitly means that `self.producer` has to be `sns_producer`; but that assumption was already there, since we were assuming that the message we get from `self.producer.create_message` had a `topic_arns` attribute.

microcosm_pubsub/producer.py
```diff
@@ -73,10 +59,17 @@ def produce(self, media_type, dct=None, uri=None, **kwargs):
 
         if self.skip:
             return
-        pubsub_message = self.create_message(media_type, dct, uri, **kwargs)
+
+        if pubsub_message is None:
+            pubsub_message = self.create_message(media_type, dct, uri, **kwargs)
```
afallou: This ensure backwards-compatibility.

afallou:
Closing this for now - it remains a valid change, but the need isn't as pressing.

PR #227, opened 2020-01-25
Add media type field in message result

afallou:
Add the message `media_type` to the `extra` dictionary when logging the result, so as to get it as a searchable field in the logged json.

### DIFF
--- a/microcosm_pubsub/tests/test_result.py
+++ b/microcosm_pubsub/tests/test_result.py
@@ -204,6 +204,7 @@ class TestMessageHandlingResult:
         ).resolve(self.message)
 
         result.extra = dict(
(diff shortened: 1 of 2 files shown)


microcosm_pubsub/tests/test_result.py
```diff
@@ -204,12 +204,13 @@ def handler(message):
         ).resolve(self.message)
 
         result.extra = dict(
+            media_type="foo.created",
```
simpleigh-globality: Not sure this tests the change above?

afallou: It tests that the merge also works for media type.

PR #228, opened 2020-01-27
Don't validate against message schema on serialization

afallou:
**Why?**
`schema.dump` in Marshmallow doesn't run any validation of the data against the schema. The stated rationale for this is
"I need to validate data from the outside world, but not from my own data" (see this [GitHub issue](https://github.com/marshmallow-code/marshmallow/issues/682#issuecomment-332758061)).

In the past we decided to bridge that gap by explictly calling `schema.validate`. However, that is awkward in two ways:
- We don't do that for the API layer (see `microcosm-flask`); which makes our approach inconsistent.
- The validation code in Marshmallow is oriented towards deserialization (follow from [this line](https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/schema.py#L794). This means that any field with `dump_only` is ignored in the schema when validating. As a consequence, it's just not possible to properly validate on serialization without working around Marshmallow conventions.

I propose making `microcosm-pubsub` consistent with Marshmallow and `microcosm-flask`, and drop validation on serializing the message.

**What?**
Remove call to `schema.validate` in coded `encode` method.

### DIFF
--- a/microcosm_pubsub/codecs.py
+++ b/microcosm_pubsub/codecs.py
@@ -79,12 +79,8 @@ class PubSubMessageCodec:
         message = dct.copy() if dct else dict()
         message.update(kwargs)
 
+        # Note that `dump` doesn't perform any validation
         dumped = self.schema.dump(message)
-
-        # NB: Schema dump doesn't run validation
-        errors = self.schema.validate(dumped)
-        if errors:
-            raise ValidationError(message=errors)
 
         # Dump to string
         return dumps(dumped)
--- a/microcosm_pubsub/tests/test_codecs.py
+++ b/microcosm_pubsub/tests/test_codecs.py
@@ -53,16 +53,6 @@ def test_encode():
     })))
 
 
-def test_encode_missing_field():
-    """
-    An invalid message will raise errors.
-
-    """
-    graph = create_object_graph("example", testing=True)
-    codec = graph.pubsub_message_schema_registry.find(DerivedSchema.MEDIA_TYPE)
-    assert_that(calling(codec.encode), raises(ValidationError))
-
-
 def test_decode():
     """
     A message will be decoded according to its schema.


rh-glob:
👍 

If there is an error in data/schema it's probably better it blows up in handlers so that it can end up in DLQ to be replayed once issue is resolved. What happens when Validation blows here atm, does it drop that data from message and leave handler to resolve?

afallou:
Right now, "it depends". With synchronous publishing, the endpoint would return a `500` and nothing would happen. With async publishing, the endpoint would return fine but the message wouldn't get published.

GabeLambda:
In the ideal world, I'd prefer if we didn't introduce the ability to publish invalid data. That said, if we're looking for symmetry between implementations, this at least makes expectations align. I'm just not sure if that's the right direction.

afallou:
More precisions on this decision:
I followed the code in Marshmallow; while the schema does have a `validate` method the call chain is:
- `validate` > `_do_load` ([here](https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/schema.py#L794))
- `_do_load` > `_deserialize` ([here](https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/schema.py#L847))
- And then `_deserialize` only takes into account the `load_fields` i.e. those that don’t have `dump_only=True` ([here](https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/schema.py#L636))
My conclusion there is that philosophy-wise Marshmallow is just not meant to support validation on dump (and this is echoed by the PR comment I linked to above).

If we ignore the philosophy for a second, practically speaking validation won't error if the schema uses `unknown=IGNORE`; but then all `dump_only` fields won’t get validated. And validating half the fields seems more confusing than not validating at all. That’s why I chose the direction of removing that validation.

pankajghosh:
> Right now, "it depends". With synchronous publishing, the endpoint would return a `500` and nothing would happen. With async publishing, the endpoint would return fine but the message wouldn't get published.

How would we be able to recover from this state?

PR #233, opened 2020-01-30
Instantiate schemas with unknown=EXCLUDE

afallou:
**Why?**
Validating schemas on serialization using Marshmallow presents
some issues, due to the fact that it is not in line with Marshmallow's
philosophy (see this [GitHub
issue](https://github.com/marshmallow-code/marshmallow/issues/682)). The
main issue is that a schema's `validate` method ignores `dump_only` fields,
so those are treated as unknown fields. However, validation is still useful
in highlighting fields that should be marked as nullable in the schema.

**What?**
- Bring back schema validation on serialization.
- Use `unknown=EXCLUDE` on schema instantiation, meaning that unknown
fields won't trigger errors.

### DIFF
--- a/microcosm_pubsub/codecs.py
+++ b/microcosm_pubsub/codecs.py
@@ -79,8 +79,12 @@ class PubSubMessageCodec:
         message = dct.copy() if dct else dict()
         message.update(kwargs)
 
-        # Note that `dump` doesn't perform any validation
         dumped = self.schema.dump(message)
+
+        # NB: Schema dump doesn't run validation
+        errors = self.schema.validate(dumped)
+        if errors:
+            raise ValidationError(message=errors)
 
         # Dump to string
         return dumps(dumped)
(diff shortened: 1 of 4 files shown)


microcosm_pubsub/codecs.py
```diff
@@ -79,9 +79,13 @@ def encode(self, dct: Optional[Dict[str, Any]] = None, **kwargs) -> str:
         message = dct.copy() if dct else dict()
         message.update(kwargs)
 
-        # Note that `dump` doesn't perform any validation
         dumped = self.schema.dump(message)
 
+        # NB: Schema dump doesn't run validation
+        errors = self.schema.validate(dumped)
+        if errors:
+            raise ValidationError(message=errors)
```
afallou: Fields that are actually not in the schema will just be ignored by `dump` (since it doesn't validate), meaning that will be absent from `dumped`. So the only unknown fields at this point will be `dump_only` fields, that `validate` doesn't take into account in the schema.

cdm-ium:
I feel like you're changing your mind here. Is the main reason because there are some errors that we can catch with this that you think are useful? I feel like I need a concrete example to make sense of this.

PR #234, opened 2020-01-31
Release 2.17.0

afallou:
Releases https://github.com/globality-corp/microcosm-pubsub/pull/233. See there for context.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.14.0
+current_version = 2.17.0
 commit = False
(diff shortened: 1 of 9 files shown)


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.14.0
+current_version = 2.17.0
```
GabeLambda: I'm guess we haven't been merging back recently.

afallou: We have, but I don't think this file was bumped; only `setup.py`

PR #236, opened 2020-05-03
Feature/devops 1413 read sqs event from json

andreyshuster:
Part of implementing Serverless Daemons feature
https://globality.atlassian.net/wiki/spaces/D/pages/1199866837/Tech+Design+Running+Daemons+with+AWS+Lambda+Serverless

Add an optional argument to ConsumerDaemon.
Add SQSJsonReader to handle messages from SQS

### DIFF
--- a/microcosm_pubsub/daemon.py
+++ b/microcosm_pubsub/daemon.py
@@ -12,10 +12,10 @@ from microcosm_pubsub.envelope import NaiveSQSEnvelope, SQSEnvelope
 
 class ConsumerDaemon(Daemon):
 
-    def __init__(self, event={}):
+    def __init__(self, event=None):
         super().__init__()
         self.bound_handlers = None
-        self.sqs_event = event
+        self.sqs_event = event or {}
 
     def make_arg_parser(self):
         parser = super().make_arg_parser()
--- a/microcosm_pubsub/daemon.py
+++ b/microcosm_pubsub/daemon.py
@@ -12,9 +12,10 @@ from microcosm_pubsub.envelope import NaiveSQSEnvelope, SQSEnvelope
 
 class ConsumerDaemon(Daemon):
 
-    def __init__(self):
+    def __init__(self, event={}):
         super().__init__()
         self.bound_handlers = None
+        self.sqs_event = event
 
     def make_arg_parser(self):
         parser = super().make_arg_parser()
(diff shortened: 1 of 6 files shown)


microcosm_pubsub/daemon.py
```diff
@@ -12,9 +12,10 @@
 
 class ConsumerDaemon(Daemon):
 
-    def __init__(self):
+    def __init__(self, event={}):
```
afallou, line 15: Using a mutable default value for an argument is an antipattern (see [explanation](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)). The simplest is to use `None` and update line 18 to 
```self.sqs_event = event or {}```

File: microcosm_pubsub/daemon.py [hunk 2]
```diff
@@ -68,7 +71,24 @@ def defaults(self):
...
+        # for AWS Lambda
+        # SQS message comes as function argument
+        # e.g
+        # ```
+        # def handler(event, context):
+        #     # processing event
+        # ```
+        # we don't use command line args here
+        if self.sqs_event:
+            config.update(
```
afallou, line 83: I'm not a fan of this, but I can't say more since it's unclear how all of this will come together.

afallou:
@andreyshuster you need to add unit-tests, they will also help clarify what your intent is here.

microcosm_pubsub/daemon.py
andreyshuster: Thanks!

andreyshuster:
@afallou thanks for help, tests added

microcosm_pubsub/tests/test_consumer.py
```diff
@@ -100,3 +112,35 @@ def test_raw_consume():
...
         media_type=DerivedSchema.MEDIA_TYPE,
     ))))
+
+
+def test_json_consume():
+    """
+    Test that message sent as JSON could be delivered
+
+    """
+    graph = create_daemon_with_naive_cache()
```
afallou, line 122: One small improvement you could make here is to pass the message in – it won't properly set the sqs_client, but it should still hit the `config.update` portion. Then you can test that the config for `sqs_envelope` and `sqs_consumer` are as expected.

andreyshuster: Thanks! Adding.

andreyshuster: It looks like adding such improvement requires adding changes to `microcosm_daemon` as well.
Creating a follow-up ticket for that.
https://globality.atlassian.net/browse/DEVOPS-1495

PR #237, opened 2020-05-15
Add support for sentry to microcosm_pubsub

rh-glob:
Why?
- easier debugging of DLQ issues
- gather data on common error types
- persistent, errors are retained after DLQ is cleared

What?
- Add ability for sentry dsn to be configured and used by a pubsub consumer to report errors 
- scrub data before sending to sentry, anything that isn't a `uuid` or known type such as `self`, `cls`, `<function` has there values redacted before sending to sentry

Questions:
- data scrubbing. More/less
- aligning on metadata to populate. Easier tracing of how an error occurs across projects.

### DIFF
--- a/microcosm_pubsub/dispatcher.py
+++ b/microcosm_pubsub/dispatcher.py
@@ -61,6 +61,7 @@ class SQSMessageDispatcher:
         self.send_metrics = graph.pubsub_send_metrics
         self.send_batch_metrics = graph.pubsub_send_batch_metrics
         self.max_processing_attempts = graph.config.sqs_message_dispatcher.message_max_processing_attempts
+        self.sentry_enabled = graph.sentry_logging.enabled
 
     def handle_batch(self, bound_handlers) -> List[MessageHandlingResult]:
         """
@@ -127,6 +128,9 @@ class SQSMessageDispatcher:
             instance.log(
                 logger=self.choose_logger(handler),
                 opaque=self.opaque,
+            )
+            instance.error_reporting(
+                self.sentry_enabled, self.opaque,
             )
             instance.resolve(message)
             return instance
--- a/microcosm_pubsub/result.py
+++ b/microcosm_pubsub/result.py
@@ -153,6 +153,18 @@ class MessageHandlingResult:
             },
         )
 
+    def error_reporting(self, sentry_enabled, opaque):
+        if not sentry_enabled:
+            return
+        from sentry_sdk import capture_exception
+        from sentry_sdk import configure_scope
+        opaque = opaque.as_dict()
+        with configure_scope() as scope:
+            scope.set_tag("x-request-id", opaque.get("X-Request-Id"))
+            scope.set_tag("message-id", opaque.get("message_id"))
+            scope.set_tag("media-type", opaque.get("media_type"))
+            capture_exception(self.exc_info)
+
     def resolve(self, message):
         if self.result.retry:
             message.nack(self.retry_timeout_seconds)
--- a/microcosm_pubsub/sentry.py
+++ b/microcosm_pubsub/sentry.py
@@ -1,0 +1,68 @@
+from dataclasses import dataclass
+from os import environ
+from typing import Any, Optional
+from unittest.mock import MagicMock
+
+from microcosm.decorators import defaults
+from microcosm.object_graph import ObjectGraph
+from microcosm_logging.decorators import logger
+from sentry_sdk.utils import BadDsn
+
+
+@dataclass
+class SentryConfig:
+    dsn: Optional[str] = None
+    enabled: bool = False
+    client: Optional[Any] = None
+
+
+def before_send(event, hint):
+    """
+    Before sending a event to sentry scrub all values that are non-uuid data from the event
+
+    """
+    for exception in event["exception"]["values"]:
+        for frame in exception["stacktrace"]["frames"]:
+            for var_name, value in frame["vars"].items():
+                if var_name in ["self", "cls"]:
+                    continue
+                elif isinstance(value, str):
+                    if value.startswith("<function "):
+                        continue
+                    elif var_name.endswith("_id"):
+                        continue
+                frame["vars"][var_name] = "<redacted>"
+    return event
+
+
+@defaults(
+    enabled=False,
+    dsn=None,
+)
+@logger
+def configure_sentry(graph: ObjectGraph):
+    enabled, dsn, sentry = False, None, None
+    if graph.config.sentry_logging.enabled and graph.config.sentry_logging.dsn:
+        enabled = graph.config.sentry_logging.enabled
+        dsn = graph.config.sentry_logging.dsn
+        if graph.metadata.testing:
+            sentry = MagicMock()
+        else:
+            import sentry_sdk
+            try:
+                sentry = sentry_sdk.init(
+                    graph.config.sentry_logging.dsn,
+                    server_name=f"{graph.metadata.name}_daemon",
+                    release=graph.config.build_info_convention.sha1,
+                    environment=environ.get("MICROCOSM_ENVIRONMENT") or "undefined",
+                    before_send=before_send,
+                )
+            except BadDsn:
+                enabled, dsn, sentry = False, None, None
+                configure_sentry.logger.error("Bad DSN value set")
+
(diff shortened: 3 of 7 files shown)


microcosm_pubsub/result.py
```diff
@@ -153,6 +153,18 @@ def log(self, logger, opaque):
...
+    def error_reporting(self, sentry_enabled, opaque):
+        opaque = opaque.as_dict()
+        if not sentry_enabled:
+            return
+        from sentry_sdk import capture_exception
+        from sentry_sdk import configure_scope
+        with configure_scope() as scope:
+            scope.set_tag("x-request-id", opaque.get("X-Request-Id"))
+            scope.set_tag("message-id", opaque.get("message_id"))
+            scope.set_tag("media-type", opaque.get("media_type"))
```
tonybajan-globality, line 165: ideally, we'd set scope when we *start* handing as message, so these tags would be available for (eg) error-level logs. but not for this PR.

microcosm_pubsub/sentry.py
```diff
@@ -0,0 +1,60 @@
...
+            for var_name, value in frame["vars"].items():
+                if var_name in ["self", "cls"]:
+                    continue
+                elif isinstance(value, str):
+                    if value.startswith("<function "):
+                        continue
+                    elif var_name.endswith("_id"):
+                        continue
+                frame["vars"][var_name] = "<redacted>"
+    return event
```
tonybajan-globality, line 34: for the record, I think this is too aggressive, but it does pretty much match what phoenix is doing [here](https://github.com/globality-corp/phoenix/blob/d08cf6e6279470c459aa9e8398cf75d693359246/src/modules/core/services/apollo-client-controller/createErrorHandler.js#L23-L34)

microcosm_pubsub/tests/test_sentry.py
```diff
@@ -0,0 +1,83 @@
...
+
+    assert_that(
+        graph.sentry_logging.enabled,
+        is_(equal_to(False))
+    )
+
+
+def test_before_send_data_removal():
+    filtered_event = before_send(sample_event, None)
+    filterd_stacktrace = filtered_event["exception"]["values"][0]["stacktrace"]["frames"]
```
tonybajan-globality, line 63: typo: filtered.

microcosm_pubsub/tests/test_sentry.py
```diff
@@ -0,0 +1,83 @@
...
+
+    assert_that(
+        filterd_stacktrace[0]["vars"],
+        is_(equal_to(
+            {'cls': "<class 'microcosm_pubsub.result.MessageHandlingResult'>",
+             'handler': '<function context_logger.<locals>.wrapped at 0x109422ef0>',
+             'message': '<redacted>'}
```
tonybajan-globality, line 70: for a test of redaction, I'd probably go for a simplified dict rather the big blob snapshot you've got.

microcosm_pubsub/sentry.py
rh-glob: Agreed, I have scaled back and included some cases that I view as safe in a python codebase context that are useful for debugging. Think this will be an ongoing process, potentially we should make this overridable by the individual services so we can easier whitelist safe data sources per service. Though that would be a future piece if PoC on a single service goes well.

microcosm_pubsub/tests/test_sentry.py
rh-glob: I started with that but the big blob test actually helped in resolving an issue (if `before_send` fails sentry error doesn't send and doesn't even show a console log error). Will see if I can remove some more of the cruft to make it more readable.

microcosm_pubsub/result.py
```diff
@@ -153,6 +153,18 @@ def log(self, logger, opaque):
             },
         )
 
+    def error_reporting(self, sentry_enabled, opaque):
+        opaque = opaque.as_dict()
```
GabeLambda, line 157: Could move this a bit further down (e.g. below the imports).

microcosm_pubsub/dispatcher.py
```diff
@@ -61,6 +61,7 @@ def __init__(self, graph):
         self.send_metrics = graph.pubsub_send_metrics
         self.send_batch_metrics = graph.pubsub_send_batch_metrics
         self.max_processing_attempts = graph.config.sqs_message_dispatcher.message_max_processing_attempts
+        self.sentry_enabled = graph.sentry_logging.enabled
```
GabeLambda, line 64: Should this be config on the dispatcher, or more generic like this? Would we push sentry logs outside of this?

microcosm_pubsub/result.py
```diff
@@ -153,6 +153,18 @@ def log(self, logger, opaque):
...
+        if not sentry_enabled:
+            return
+        from sentry_sdk import capture_exception
+        from sentry_sdk import configure_scope
```
GabeLambda, line 161: If these hadn't been installed, but sentry had been enabled, would you want something friendlier than just raising a bare `ImportError`?

microcosm_pubsub/sentry.py
```diff
@@ -0,0 +1,60 @@
...
+    for exception in event["exception"]["values"]:
+        for frame in exception["stacktrace"]["frames"]:
+            for var_name, value in frame["vars"].items():
+                if var_name in ["self", "cls"]:
+                    continue
+                elif isinstance(value, str):
+                    if value.startswith("<function "):
+                        continue
+                    elif var_name.endswith("_id"):
+                        continue
```
GabeLambda, line 32: Maybe:
```
if all([
    value.startswith("<function "),
    var_name.endswith("_id"),
]):
    continue
```

microcosm_pubsub/tests/test_sentry.py
```diff
@@ -0,0 +1,83 @@
...
+            ),
+            build_info_convention=dict(
+                sha1="some-sha",
+            )
+        )
+
+    graph = create_object_graph("example", testing=False, loader=loader)
+    try:
+        graph.sentry_logging.enabled
+        assert False
```
GabeLambda, line 28: I think this is a bit more...expressive?
`assert_that(calling(lambda: graph.sentry_logging.enabled), raises(BadDsn))`

microcosm_pubsub/dispatcher.py
rh-glob: Potentially, sentry is moving in a direction of being more than just error reporting, so you push any log like data that is relevant and tag it accordingly so you can quickly find traces/related data.

Also my thinking was keeping the logic for deciding if sentry is enabled quite generic so that if this goes well. The logic for checking if sentry is enabled and adding it to the graph and the cleaning functions moves to it's own library that would be used by both pubsub/flask libs.

microcosm_pubsub/result.py
rh-glob: The package would be installed it's not optional in the `setup.py`. So we wouldn't have an `ImportError`, the other error that we could have is a missing dsn but that is protected against as setup only succeeds if both enabled flag and dsn is set. Will add some extra handling so if a bad url is setup the sentry is just disabled too.

microcosm_pubsub/sentry.py
rh-glob: `any`, but also don't think this adds much to readability. Tend to avoid any/all as it evaluates all elements which in some cases is a waste losing the benefit of the early exit.

microcosm_pubsub/tests/test_sentry.py
rh-glob: Oddly enough that wasn't working as it's what I tried first. Updating the test to handle `BadDsn` and disable sentry in this case.

microcosm_pubsub/result.py
GabeLambda: So, we tend to frown a bit on introducing new 3rd-party libs if they aren't central to the functioning of a given lib (idea being that this is open-source, and don't want to push the dep on all downstream consumers). (Yes, I know this isn't widely-used. But it's beside the point).

Also, why not put this as a top-level import if it should be available globally?

rh-glob: Fair criticism, I will look to action this if internal POC goes well and pull this out into a separate lib and make it optional within this package.

PR #240, opened 2020-05-22
Only report Error state results to sentry

rh-glob:
Why?
- Handler can exit early with more cases than just error, such as Nack, SkipMessage this was generating events in sentry that were not beneficial

What?
- Only report error to sentry if result_type is `FAILED` or `EXPIRED`
- Only report error if there is an `exc_info` present

### DIFF
--- a/microcosm_pubsub/result.py
+++ b/microcosm_pubsub/result.py
@@ -154,13 +154,13 @@ class MessageHandlingResult:
         )
 
     def error_reporting(self, sentry_enabled, opaque):
-        if any([
-            not sentry_enabled,
-            self.result not in [
+        if not all([
+            sentry_enabled,
+            self.result in [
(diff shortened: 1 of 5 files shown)


microcosm_pubsub/result.py
```diff
@@ -154,8 +154,18 @@ def log(self, logger, opaque):
...
     def error_reporting(self, sentry_enabled, opaque):
-        if not sentry_enabled:
+        if any([
+            not sentry_enabled,
+            self.result not in [
+                MessageHandlingResultType.FAILED,
+                MessageHandlingResultType.EXPIRED,
+            ],
+            not self.exc_info,
+        ]):
```
tonybajan-globality, line 164: I think this'd be clearer if you demorgan out the `not` to a `if not all([...])` instead.

PR #241, opened 2020-05-22
Make sentry an optional dependency

rh-glob:
Why?
- We shouldn't force library users to install libraries that aren't essential to the library

What?
- Move sentry to `extras_require`
- Handle import error when initializing sentry


GabeLambda:
I think you may either want to skip the test, or update the build to install the extras.

PR #242, opened 2020-05-26
Release/2.19.0

rh-glob:
Release:
- Only report Error states to sentry ignore Nack/SkipMessage:
https://github.com/globality-corp/microcosm-pubsub/pull/240
- Make sentry an optional dependency:
https://github.com/globality-corp/microcosm-pubsub/pull/241


microcosm_pubsub/result.py
```diff
@@ -154,8 +154,18 @@ def log(self, logger, opaque):
         )
 
     def error_reporting(self, sentry_enabled, opaque):
-        if not sentry_enabled:
+        if not all([
+            sentry_enabled,
+            self.result in [
+                MessageHandlingResultType.FAILED,
+                MessageHandlingResultType.EXPIRED,
```
afallou, line 161: Casing this late in the flow seems weird. What's the reason?

rh-glob: Not sure what you mean by casing. In terms of position in flow, these result types should behave as normal for everything else in this class. It's only the sentry integration that doesn't care about the other result types.

PR #244, opened 2020-05-28
Fixing SQSJsonReader tests. 

andreyshuster:
Making sure, it has correct Consumer and Envelope types
https://globality.atlassian.net/browse/DEVOPS-1495
This PR adds tests which validate code for a SQSJsonReader, including envelope and consumer.
All other test suites work with MagicMock consumer.


microcosm_pubsub/daemon.py
```diff
@@ -82,7 +82,7 @@ def defaults(self):
         if self.sqs_event:
             config.update(
                 sqs_envelope=dict(
-                    strategy_name=NaiveSQSEnvelope.__name__,
+                    strategy_name=RawSQSEnvelope.__name__,
```
andreyshuster, line 85: According to documentation in `envelope.py` module, `NaiveSQSEnvelope` used for debugging

microcosm_pubsub/tests/test_consumer.py
```diff
@@ -119,28 +125,29 @@ def test_json_consume():
...
+    assert_that(graph.sqs_consumer, is_(instance_of(SQSConsumer)))
+    assert_that(graph.sqs_envelope, is_(instance_of(RawSQSEnvelope)))
     assert_that(messages, has_length(1))
     assert_that(messages[0].consumer, is_(equal_to(graph.sqs_consumer)))
-    assert_that(messages[0].message_id, is_(equal_to(MESSAGE_ID)))
     assert_that(messages[0].receipt_handle, is_(equal_to(RECEIPT_HANDLE)))
     assert_that(messages[0].content, is_(equal_to(dict(
         data="data",
-        media_type=DerivedSchema.MEDIA_TYPE,
+        mediaType=DerivedSchema.MEDIA_TYPE,
```
rh-glob, line 152: Is this casing change due to the envelope change?

rh-glob: The worry would be when we switch default behaviour to the previous lambda type with `media_type` any of them that had any hard-coded assumptions about the message structure would break. Consider it breaking a contract in the message structure that something might rely on.

From a quick search of our consumers this doesn't seem to be the case, the only place we directly refer to `mediaType` is in this library and tests. Though that wouldn't be ideal if switching to the new consumer type broke 100s of tests across a lot of repos ;)

microcosm_pubsub/daemon.py
```diff
@@ -82,7 +82,7 @@ def defaults(self):
         if self.sqs_event:
             config.update(
                 sqs_envelope=dict(
-                    strategy_name=NaiveSQSEnvelope.__name__,
+                    strategy_name=CodecSQSEnvelope.__name__,
```
simpleigh-globality, line 85: Is this backwards-compatible with a daemon deployed as currently? If not then this will make deployment to Lambda more complicated and risk breaking existing daemons if they upgrade the library version. Not necessarily a problem, but:

* if we can maintain backwards-compatibility then we should
* if we can't then we need a major version number bump and some more planning 😄

rh-glob: `sqs_event` won't be set on any daemons currently deployed other than the daemons that @andreyshuster is testing with. The pre-cursor of this work is already present in the release that added sentry support. 

https://github.com/globality-corp/microcosm-pubsub/pull/238/files#diff-063fc687b977ac99757488ad2741ee96R82

simpleigh-globality: Thanks. With hindsight `sqs_event` could have been clearer (at least for this 🐻 of very little brain) 😄.

microcosm_pubsub/tests/test_consumer.py
```diff
@@ -119,26 +125,27 @@ def test_json_consume():
...
+    # simulate the response structure
+    # replacing MagicMock with real reader
+    graph.sqs_consumer.sqs_client = SQSJsonReader(event)
     messages = graph.sqs_consumer.consume()
 
     # and response translated properly
+    assert_that(graph.sqs_consumer, is_(instance_of(SQSConsumer)))
+    assert_that(graph.sqs_envelope, is_(instance_of(CodecSQSEnvelope)))
     assert_that(messages, has_length(1))
     assert_that(messages[0].consumer, is_(equal_to(graph.sqs_consumer)))
```
afallou, line 148: Isn't there another assertion you could add checking that the config is as expected? IIRC when instantiating the daemon you're adding config for the Json case.

PR #249, opened 2020-06-19
Release/2.21.0

rh-glob:
Releases https://github.com/globality-corp/microcosm-pubsub/pull/248 - Change sentry graph registration point


rh-glob:
> We generally follow the rules of semantic versioning, but a good rule of thumb is to bump minor by default, major for significant changes, patch for hotfixes.

PR #251, opened 2020-07-07
Add common lambda handler code to library

andreyshuster:
Add AWS Lambda daemon handler to library
See context here  https://github.com/globality-corp/uhura/pull/64
@tonybajan-globality @afallou  @rh-glob 
Please suggest how it could be tested better.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ setup(
         "microcosm-caching>=0.2.0",
         "microcosm-daemon>=1.0.0",
         "microcosm-logging>=1.3.0",
+        "isort<5.0.0",
     ],
     extras_require={
         "metrics": "microcosm-metrics>=2.5.0",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -26,6 +26,7 @@
         "microcosm-caching>=0.2.0",
         "microcosm-daemon>=1.0.0",
         "microcosm-logging>=1.3.0",
+        "isort<5.0.0",
```
rh-glob, line 29: Is this required for build to pass? Ideally we wouldn't have such a restriction in a library, as it makes it more complicated to to upgrade isort in the future for the services using this library.

andreyshuster: Unfortunately. 
`Lint` job fails because of that.
We have this issue in all repositories at this moment.

andreyshuster: Just saw Tony's comment in Slack on that. let me try to kick the build again and see what happens

andreyshuster: Looks like it's not required anymore!

PR #253, opened 2020-07-14
Update lambda handler code. 

andreyshuster:
Update tests. Create new SQS enveloping strategy for daemons

@rh-glob @afallou @tonybajan-globality Please, review!

For some reason, SQS queue sends a message to a Lambda function in a slightly different format, where message dictionary keys are not capitalized. 
So I decided to create a separate class to handle this. 
This change also allowed me to pass the message to handler without additional json operations on it.

### DIFF
--- a/microcosm_pubsub/daemon.py
+++ b/microcosm_pubsub/daemon.py
@@ -2,8 +2,6 @@
 Consume Daemon main.
 
 """
-from json import loads
-
 from microcosm.loaders import load_each, load_from_dict
 from microcosm_daemon.api import SleepNow
 from microcosm_daemon.daemon import Daemon
--- a/microcosm_pubsub/daemon.py
+++ b/microcosm_pubsub/daemon.py
@@ -9,7 +9,7 @@ from microcosm_daemon.api import SleepNow
 from microcosm_daemon.daemon import Daemon
 
 from microcosm_pubsub.consumer import STDIN
-from microcosm_pubsub.envelope import CodecSQSEnvelope, NaiveSQSEnvelope, SQSEnvelope
+from microcosm_pubsub.envelope import LambdaSQSEnvelope, NaiveSQSEnvelope, SQSEnvelope
 
 
 class ConsumerDaemon(Daemon):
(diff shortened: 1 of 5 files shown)


microcosm_pubsub/tests/test_consumer.py
```diff
@@ -126,10 +126,10 @@ def test_json_consume():
 
     """
```
rh-glob, line 127: nit: Can you update this tests docstring and/or name is callout the lambda style use case this is testing.

andreyshuster: sure, moment

microcosm_pubsub/envelope.py
```diff
@@ -210,6 +210,36 @@ def parse_body(self, raw_message):
...
+        """
+        Validate the message body.
+
+        Just checks for tampering; schema validation occurs once we know the type of message.
+
+        """
+        expected_md5_of_body = raw_message["md5OfBody"]
+        actual_md5_of_body = md5(body).hexdigest()
+        if expected_md5_of_body != actual_md5_of_body:
+            raise Exception("MD5 validation failed. Expected: {} Actual: {}".format(
```
rh-glob, line 237: A more specific exception class would be good, potentially a `ValidationError` as done in `codecs.py`

andreyshuster: Ok, sure. I took this code as-is from SQSEnvelope base class.

microcosm_pubsub/daemon.py
```diff
@@ -64,15 +62,10 @@ def handler(event, context):
             if "warm" in event:
                 return "warming up"
 
-            message = event["Records"][0]["body"]
```
rh-glob, line 67: I don't see this change in event contract reflected in the test. Test modified doesn't seem to have `Records`.

andreyshuster: you right

andreyshuster: This is true. The test does not have Records because it's testing only the consumer itself.

PR #259, opened 2021-01-21
Release 2.23.0

adamhadani:
- Incorporate changes from Nov. 2020 to remove tornado<6 dependency/version pin and bump microcosm dep to >3.0
- pin isort at <5
- incorporate changes to to some daemon / lambda handling logic by @pankajghosh

### DIFF
--- a/microcosm_pubsub/daemon.py
+++ b/microcosm_pubsub/daemon.py
@@ -2,6 +2,8 @@
 Consume Daemon main.
 
 """
+import argparse
+
 from microcosm.loaders import load_each, load_from_dict
 from microcosm_daemon.api import SleepNow
 from microcosm_daemon.daemon import Daemon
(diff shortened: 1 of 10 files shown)


microcosm_pubsub/daemon.py
```diff
@@ -46,27 +54,36 @@ def process(self):
...
+            # microcosm graph loading time would slow the message processing
             if "warm" in event:
+                daemon.graph.logger.debug("Processing warm-up cloudwatch event")
                 return "warming up"
 
             # we configure SQS Queue to use batches of 1,
             # so received event contains
             # another stringified json with actual message inside
-            daemon = cls(event=event["Records"][0])
+            daemon.graph.logger.debug("Processing SQS event: {}".format(event["Records"][0]))
```
adamhadani, line 82: @pankajghosh do we want this debug log line here? it also violates some [indirection principles](https://en.wikipedia.org/wiki/Law_of_Demeter) given how deep it goes into the object graph..

microcosm_pubsub/daemon.py
```diff
@@ -46,27 +54,36 @@ def process(self):
...
+            daemon.graph.sqs_consumer.set_sqs_event(event["Records"][0])
```
adamhadani, line 83: is `event["Records"]` always guaranteed to be non-None / have at least one member @pankajghosh ?

PR #256, opened 2020-10-04
modified daemon handler to initialize at startup, and optimize warm startup times

pankajghosh:
Closes https://globality.atlassian.net/browse/DEVOPS-2542

PR #260, opened 2021-01-21
Revert "modified daemon handler to initialize microcosm graph at startup, and optimize lambda warm-startup times (#256)"

adamhadani:
This reverts commit c98a188fcd5b3f358c7171dae0c39a33c5774a4e. This project has been abandoned.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -27,7 +27,7 @@ if [ "$1" = "test" ]; then
 elif [ "$1" = "lint" ]; then
     # Install standard linting dependencies; YMMV
     pip --quiet install \
-        .[lint] flake8 "isort<5.0.0" flake8-print flake8-logging-format flake8-isort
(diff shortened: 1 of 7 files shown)


GabeLambda (approved):

entrypoint.sh
```diff
@@ -27,7 +27,7 @@ if [ "$1" = "test" ]; then
 elif [ "$1" = "lint" ]; then
     # Install standard linting dependencies; YMMV
     pip --quiet install \
-        .[lint] flake8 "isort<5.0.0" flake8-print flake8-logging-format flake8-isort
+        .[lint] "isort<5.0.0" flake8 flake8-print flake8-logging-format flake8-isort
```
adamhadani, line 30: I re-introduced the isort pin after reverting this commit as still needed - no change here just ordering

PR #264, opened 2021-09-03
Add optional tracing with Dynatrace OneAgent

tonybajan-globality:
If the Dynatrace OneAgent SDK is installed, instrument messages sent
and received.

If it is not installed, silently do nothing instead.

### DIFF
--- a/microcosm_pubsub/consumer.py
+++ b/microcosm_pubsub/consumer.py
@@ -5,7 +5,7 @@ Message consumer.
 from os.path import exists
 from urllib.parse import urlparse
 
-from boto3 import Session
+from boto3.session import Session
 from microcosm.api import defaults, typed
 from microcosm_logging.decorators import logger
 
--- a/microcosm_pubsub/producer.py
+++ b/microcosm_pubsub/producer.py
@@ -10,7 +10,7 @@ from logging import Logger
 from time import time
 from typing import Dict, List
 
-from boto3 import Session
+from boto3.session import Session
 from botocore.client import Config
 from microcosm.api import defaults, typed
 from microcosm.errors import NotBoundError
@@ -21,6 +21,7 @@ from microcosm_pubsub.batch import MessageBatchSchema
 from microcosm_pubsub.constants import PUBLISHED_KEY
 from microcosm_pubsub.conventions.naming import make_media_type
 from microcosm_pubsub.errors import TopicNotDefinedError
+from microcosm_pubsub.tracing import add_trace_to_message, trace_outgoing_message
 
 
 @dataclass
@@ -113,13 +114,18 @@ class SNSProducer:
         publish_result = "SUCCESS"
         publish_exception = None
 
-        with elapsed_time(extra):
+        with trace_outgoing_message(pubsub_message.topic_arn) as tracer, elapsed_time(extra):
             try:
+                add_trace_to_message(tracer, pubsub_message)
                 result = self.sns_client.publish(
                     TopicArn=pubsub_message.topic_arn,
                     Message=pubsub_message.message,
                     MessageAttributes=pubsub_message.message_attributes,
                 )
+                tracer.set_vendor_message_id(result["MessageId"])
+                request_id = pubsub_message.opaque_data.get("X-Request-Id")
+                if request_id:
+                    tracer.set_correlation_id(request_id)
             except Exception as e:
                 publish_exception = e
                 publish_result = "FAILURE"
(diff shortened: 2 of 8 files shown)


microcosm_pubsub/producer.py
```diff
@@ -113,13 +114,18 @@ def publish_message(self, pubsub_message: PubsubMessage):
...
+        with trace_outgoing_message(pubsub_message.topic_arn) as tracer, elapsed_time(extra):
             try:
+                add_trace_to_message(tracer, pubsub_message)
                 result = self.sns_client.publish(
                     TopicArn=pubsub_message.topic_arn,
                     Message=pubsub_message.message,
                     MessageAttributes=pubsub_message.message_attributes,
                 )
+                tracer.set_vendor_message_id(result["MessageId"])
+                request_id = pubsub_message.opaque_data.get('X-Request-Id')
```
isabelcosta-globality, line 126:
```suggestion
                request_id = pubsub_message.opaque_data.get("X-Request-Id")
```
nit double quotes

microcosm_pubsub/tracing.py
```diff
@@ -0,0 +1,96 @@
...
+            message["opaqueData"][OPAQUE_TAG_KEY] = tag
+            pubsub_message.message = json.dumps(message)
+        except Exception:
+            pass
+
+
+@contextmanager
+def trace_outgoing_message(topic_arn: str):
+    if not oneagent:
+        yield StubOutgoingTracer()
```
JP-Globality, line 55: Nice way of dealing with imported / not-imported libraries 👍

JP-Globality (approved):

microcosm_pubsub/tests/test_tracing.py
```diff
@@ -0,0 +1,128 @@
...
+            MessageAttributes=dict(
+                media_type=dict(
+                    DataType='String',
+                    StringValue='application/vnd.microcosm.derived'
+                ),
+            ),
+            TopicArn='topic-name'
+        )
+        message = json.loads(graph.sns_producer.sns_client.publish.call_args[1]['Message'])
+        assert_that(message["opaqueData"]["X-Request-Id"], equal_to("req-id-1234"))
```
isabelcosta-globality, line 51: would the usual `assert` instead `assert_that` work in this file? just wondering if you did this intentional.

isabelcosta-globality: gotcha! I did feel this was intentional 👍🏾  thank you for explaining!

microcosm_pubsub/consumer.py
```diff
@@ -5,7 +5,7 @@
 from os.path import exists
 from urllib.parse import urlparse
 
-from boto3 import Session
+from boto3.session import Session
```
sethisernhagen, line 8: Why are both of these valid? Why change it?

microcosm_pubsub/tracing.py
```diff
@@ -0,0 +1,96 @@
+"""
+Support optional tracing using Dynatrace oneagent.
+
+If the dynatrace SDK is not installed, does nothing.
+
+"""
```
sethisernhagen, line 6: Should this be `tracing.py` or `dynatracing.py`

microcosm_pubsub/consumer.py
tonybajan-globality: Session is imported into `boto3.__init__.py` so it works. It's not part of the typing stubs, though, so mypy complains about the indirect import.  I guess this is due to newer stubs or newer mypy since the last changes to this library.

microcosm_pubsub/tracing.py
tonybajan-globality: I went with tracing.py as we could perhaps choose between different tracing backends here - though the API would definitely need some work.

sethisernhagen: This can be the "opinionated" part. 👍

microcosm_pubsub/tracing.py
```diff
@@ -0,0 +1,96 @@
...
+    from oneagent.common import ChannelType, MessagingDestinationType
+    from oneagent.sdk import Channel
+except ImportError:
+    oneagent = None
+    ChannelType = None
+    MessagingDestinationType = None
+    Channel = None
+
+
+AWS_SNS = "SNS"
```
sethisernhagen, line 26: Cool. Wasn't event thinking about publishing.

microcosm_pubsub/tests/test_tracing.py
```diff
@@ -0,0 +1,128 @@
...
+            str_tag="tag-5678",
+        )
+        tracer_manager = sdk.trace_incoming_message_process.return_value
+        tracer_manager.__enter__.assert_called_once_with()
+        tracer_obj = tracer_manager.__enter__.return_value
+        tracer_obj.set_vendor_message_id.assert_called_once_with(
+            "message-id-5678"
+        )
+        tracer_obj.set_correlation_id.assert_called_once_with(
+            "req-id-5678"
```
sethisernhagen, line 126: Can't wait to see the demo!

PR #277, opened 2022-08-03
Allow a reason and extras when creating result from Nack exception

kcox-globality:
Allow an optional `reason` and `extras` when creating a result from a Nack exception. This will allow us to include additional information in logs when Nacks are raised by handlers -- which become RETRIED messages.

Here's an example using `reason` (see the `log_reason` field):
Screen Shot 2022-07-25 at 5 16 40 PM
^^ from https://github.com/globality-corp/euclides/pull/463

Closes [GLOB-66164](https://globality.atlassian.net/browse/GLOB-66164)

### DIFF
--- a/microcosm_pubsub/errors.py
+++ b/microcosm_pubsub/errors.py
@@ -20,13 +20,17 @@ class Nack(Exception):
     and the SQS dead letter queue (DLQ) configuration, if any.
 
     """
-    def __init__(self, visibility_timeout_seconds, reason="", extra=None):
+    def __init__(self, visibility_timeout_seconds, reason=None, extra=None):
         super().__init__(reason)
         self.visibility_timeout_seconds = visibility_timeout_seconds
         self.extra = extra or dict()
 
     def __repr__(self):
-        return "Nack({})".format(self.visibility_timeout_seconds)
+        return "Nack(visibility_timeout_seconds={timeout}, reason={reason}, extra={extra})".format(
+            timeout=self.visibility_timeout_seconds,
+            reason=str(self),
+            extra=self.extra,
+        )
 
 
 class SkipMessage(Exception):
--- a/microcosm_pubsub/errors.py
+++ b/microcosm_pubsub/errors.py
@@ -20,8 +20,10 @@ class Nack(Exception):
     and the SQS dead letter queue (DLQ) configuration, if any.
 
     """
-    def __init__(self, visibility_timeout_seconds):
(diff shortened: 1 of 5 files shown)


jroundsglobality (approved):

mchen-globality (approved):

microcosm_pubsub/errors.py
```diff
@@ -20,8 +20,10 @@ class Nack(Exception):
     and the SQS dead letter queue (DLQ) configuration, if any.
 
     """
-    def __init__(self, visibility_timeout_seconds):
+    def __init__(self, visibility_timeout_seconds, reason="", extra=None):
```
sethisernhagen, line 23: Can reason be None instead of empty string? Then function signature would match `TTLExpired()`.

sethisernhagen: Interesting that `reason` is required for `SkipMessage`. Nack should probably be like that someday.

microcosm_pubsub/errors.py
```diff
@@ -20,8 +20,10 @@ class Nack(Exception):
...
+        super().__init__(reason)
         self.visibility_timeout_seconds = visibility_timeout_seconds
+        self.extra = extra or dict()
```
sethisernhagen, line 26: Should we update `__repr__` to match contructor?

microcosm_pubsub/tests/test_result.py
```diff
@@ -153,6 +153,32 @@ def handler(message):
...
+                ),
+            ),
+        )
+        # nack with custom retry visibility timeout
+        self.graph.sqs_consumer.sqs_client.change_message_visibility.assert_called_with(
+            QueueUrl="queue",
+            ReceiptHandle=RECEIPT_HANDLE,
+            VisibilityTimeout=3,
+        )
+
```
sethisernhagen, line 181: Nice!

sethisernhagen (approved):
This should have been this way for a long time.

setup.py
```diff
@@ -20,6 +20,7 @@
     install_requires=[
         "boto3>=1.5.8",
         "dataclasses;python_version<'3.7'",
+        "flake8<=4.0.1"
```
tonybajan-globality, line 23: let's not pin this here. It's not a runtime dependency and will prevent all downstream dependants from upgrading at their own pace.  Once this branch is updated from master this should appear in the "lint" extra instead.

And I guess we hope the maintainers of [flake8-logging-format](https://github.com/globality-corp/flake8-logging-format) fix their builds 😳?

PR #279, opened 2022-08-22
Add Nack reasons when retrieving message URI resource

sethisernhagen:
## What?

- update uri_handler to include reason value when raising Nack

## Why?

Without the reason, determining why a handler nacked is harder than it
needs to be.


ronanm-globality (approved):

rh-glob (approved):

jpringle-globality:
cool. i didn't know about `reason="..."`.

